PR #3, opened 2022-10-25
server with DB

zinchenko-anastasiia:


src/controllers/expenses.js
```diff
@@ -0,0 +1,141 @@
...
+
+  const foundExpenses = await getExpenseById(expenseId);
+
+  if (!foundExpenses) {
+    res.sendStatus(404);
+
+    return;
+  }
+
+  const { title } = req.body;
```
NataliyaAn, line 124: Expense has some field for update "title"  "amount"  "category"  "note". You update only title

src/controllers/expenses.js
```diff
@@ -0,0 +1,141 @@
...
+const update = async(req, res) => {
+  const { expenseId } = req.params;
+
+  const foundExpenses = await getExpenseById(expenseId);
+
+  if (!foundExpenses) {
+    res.sendStatus(404);
+
+    return;
+  }
```
NataliyaAn, line 122: We have to do two errors - 400 and 404. 400 for 'Bad request' if id user was not found

src/controllers/users.js
```diff
@@ -0,0 +1,91 @@
...
+const remove = async(req, res) => {
+  const { userId } = req.params;
+
+  const foundUser = await getUserById(userId);
+
+  if (!foundUser) {
+    res.sendStatus(404);
+
+    return;
+  }
```
NataliyaAn, line 57: add check for nameUser. If name = underfind - return status 400

src/controllers/expenses.js
```diff
@@ -0,0 +1,141 @@
...
+        new Date(expense.spent_at).getTime() > new Date(from).getTime()
+        && new Date(expense.spent_at).getTime() < new Date(to).getTime()
+      )
+    );
+
+    res.send(foundExpensesByDate);
+    res.statusCode = 200;
+
+    return;
+  }
```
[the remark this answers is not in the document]
zinchenko-anastasiia: I dont change it because in the previous task we had 3 different conditions: for users, between dates and by category😊 but your comment is useful for me thx

main.js
```diff
@@ -0,0 +1,9 @@
+'use strict';
+
+const { createServer } = require('./src/createServer');
+
+createServer()
+  .listen(3000, () => {
+    // eslint-disable-next-line no-console
+    console.log('Server is running on localhost:3000');
+  });
```
dellachaise, line 9: Do you still need this code?

src/controllers/expenses.js
```diff
@@ -0,0 +1,144 @@
...
+  getExpenseById,
+  updateExpenses,
+  removeExpenses,
+  createNewExpense,
+} = require('../services/expenses');
+
+const { getUserById } = require('../services/users');
+
+const addNew = async(req, res) => {
+  const { userId, title, amount, category, note } = req.body;
```
dellachaise, line 14: You need here only `userId`.  You could make your code sporter 👇 
```suggestion
  const { userId } = req.body;
```

And in line 25 use `req.body` with the spread operator.

src/controllers/expenses.js
```diff
@@ -0,0 +1,144 @@
...
+
+  const foundExpens = await getExpenseById(expenseId);
+
+  if (!foundExpens) {
+    res.sendStatus(404);
+
+    return;
+  }
+
+  res.statusCode = 200;
```
dellachaise, line 47: This line is redundant, delete it. And delete the same redundant lines all over the code.

200 status code - is the default one.

[Read more details here](https://stackoverflow.com/questions/53403078/express-is-it-necessary-to-respond-with-status-200)

src/createServer.js
```diff
@@ -0,0 +1,21 @@
+/* eslint-disable no-shadow */
```
dellachaise, line 1: Using `eslint-disable` is a bag practice. Try to avoid it here and in test tasks. Find the way to solve it without `eslint-disable`.

dellachaise (approved):
Good job 👍 

Just need to add some improvements according to the comments. And pay attention to such cases during compliting your future test tasks.

PR #6, opened 2022-10-26
solution

Anton-Liada:


src/controllers/user.js
```diff
@@ -0,0 +1,106 @@
...
+  const users = await userService.getAll();
+
+  if (!users.length) {
+    res.send([]);
+
+    return;
+  }
+
+  res.send(users);
+};
```
VasylZinchenko, line 30:
```suggestion
const getAll = async(req, res) => {
  const users = await userService.getAll();

  if (!users.length) {
    res.send([]);

    res.send(users);
};
```

src/controllers/user.js
```diff
@@ -0,0 +1,106 @@
...
+
+const update = async(req, res) => {
+  const userId = +req.params.userId;
+  const { name } = req.body;
+
+  if (typeof userId !== 'number') {
+    res.sendStatus(400);
+
+    return;
+  }
```
VasylZinchenko, line 61: i have doubts about this check in case of update

src/controllers/user.js
```diff
@@ -0,0 +1,106 @@
...
+};
+
+const remove = async(req, res) => {
+  const userId = +req.params.userId;
+
+  if (typeof userId !== 'number') {
+    res.sendStatus(400);
+
+    return;
+  }
```
VasylZinchenko, line 86: and here

VasylZinchenko (approved):
Well done! 👍

src/routes/userRouter.js
```diff
@@ -0,0 +1,19 @@
...
+function initUserRouter(app) {
+  app.post('/', userController.add);
+
+  app.get('/', userController.getAll);
+
+  app.get('/:userId', userController.getOne);
+
+  app.patch('/:userId', userController.update);
+
+  app.delete('/:userId', userController.remove);
```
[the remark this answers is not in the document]
Anton-Liada: this function takes the router as the argument )

src/controllers/user.js
Anton-Liada: I do this check, because in my previous work without it tests failed...

PR #8, opened 2022-10-27
Develop

OlehMy:


src/services/users.js
```diff
@@ -0,0 +1,82 @@
...
+  const foundedUser = await getOne(userId);
+
+  await client.query(`
+    DELETE
+    FROM users
+    WHERE id=$1
+    `, [userId]
+  );
+
+  return foundedUser;
```
VolodymyrChuyko, line 59: The Postgres delete query returns the number of deleted rows. You can use that to check if your query was successful

src/services/expenses.js
```diff
@@ -0,0 +1,124 @@
...
+  );
+
+  const result = await client.query(`
+    SELECT *
+    FROM expenses
+    ORDER BY id DESC
+    LIMIT 1
+  `);
+
+  return result.rows[0];
```
VolodymyrChuyko, line 81:
```suggestion
  const result = await client.query(`
    INSERT INTO expenses(user_id, spent_at, title, amount, category, note)
    VALUES (${valuesString})
    `, [...values]
  );

  return result.rows[0];
```
Insert query returns created expense, so you can get rid of the redundant query.

src/services/expenses.js
```diff
@@ -0,0 +1,124 @@
...
+  await client.query(`
+    DELETE
+    FROM expenses
+    WHERE id=$1
+    `, [expenseId]
+  );
+
+  return foundedExpenses;
+}
+
```
VolodymyrChuyko, line 96: The Postgres delete query returns the number of deleted rows. You can use that to check if your query was successful

src/controllers/users.js
```diff
@@ -0,0 +1,90 @@
...
+  remove: removeUser,
+  update: updateUser,
+  normalize,
+} = require('../services/users');
+
+const getAll = async(req, res) => {
+  const users = await getAllUsers();
+
+  res.send(users.map(normalize));
+};
```
VolodymyrChuyko, line 16:
```suggestion
  res.statusCode = 200
  res.send(users.map(normalize));
};
```

src/controllers/users.js
```diff
@@ -0,0 +1,90 @@
...
+  if (isNaN(+userId)) {
+    res.sendStatus(400);
+
+    return;
+  }
+
+  const foundUser = await getOneUser(userId);
+
+  foundUser ? res.send(normalize(foundUser)) : res.sendStatus(404);
+};
```
VolodymyrChuyko, line 30: You should send status code 200 on success

src/controllers/users.js
```diff
@@ -0,0 +1,90 @@
...
+    res.sendStatus(404);
+
+    return;
+  }
+
+  const updatedUser = await updateUser(userId, name);
+
+  res.send(normalize(updatedUser));
+};
+
```
VolodymyrChuyko, line 83:
```suggestion
  const updatedUser = await updateUser(userId, name);
  
  res.statusCode = 201;
  res.send(normalize(updatedUser));
};

```

src/controllers/expenses.js
```diff
@@ -0,0 +1,144 @@
...
+  if (dateIsInvalid) {
+    res.sendStatus(400);
+
+    return;
+  }
+
+  const filteredExpenses = await getManyExpenses(userId, category, from, to);
+
+  res.send(filteredExpenses.map(normalize));
+};
```
VolodymyrChuyko, line 37:
```suggestion

  const filteredExpenses = await getManyExpenses(userId, category, from, to);
  
  res.statusCode = 200;
  res.send(filteredExpenses.map(normalize));
};
```

src/controllers/expenses.js
```diff
@@ -0,0 +1,144 @@
...
+    res.sendStatus(400);
+
+    return;
+  }
+
+  const foundExpense = await getOneExpense(expenseId);
+
+  foundExpense ? res.send(normalize(foundExpense)) : res.sendStatus(404);
+};
+
```
VolodymyrChuyko, line 52: you should send status code 200 on success

src/controllers/expenses.js
```diff
@@ -0,0 +1,144 @@
...
+  if (!foundExpense) {
+    res.sendStatus(404);
+
+    return;
+  }
+
+  const updatedExpense = await updateExpense(expenseId, req.body);
+
+  res.send(normalize(updatedExpense));
+};
```
VolodymyrChuyko, line 136:
```suggestion

  const updatedExpense = await updateExpense(expenseId, req.body);

  res.statusCode = 200;
  res.send(normalize(updatedExpense));
};
```

VolodymyrChuyko (changes requested):
Great job. Just a few small issues.

src/services/users.js
OlehMy: Good idea))

src/services/users.js
```diff
@@ -0,0 +1,82 @@
...
+
+  const result = await client.query(`
+    SELECT *
+    FROM users
+    ORDER BY id DESC
+    LIMIT 1
+  `);
+
+  return result.rows[0];
+}
```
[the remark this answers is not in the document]
OlehMy: And here the same)

src/services/expenses.js
OlehMy: And here the same))

src/controllers/users.js
OlehMy: But we have this status by default 🤔

src/controllers/expenses.js
OlehMy: But we have this status by default 🤔

src/controllers/expenses.js
OlehMy: But we have this status by default 🤔

src/controllers/expenses.js
OlehMy: But we have this status by default 🤔

OlehMy: Thanks for reviewing 😉

src/services/users.js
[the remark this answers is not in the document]
VolodymyrChuyko: Ok. Maybe it doesn't work for pg. I just thought it works similar to sequalize

src/controllers/users.js
VolodymyrChuyko: Oh, great! Thanks for your contribution to my development 🙂

src/services/expenses.js
```diff
@@ -0,0 +1,122 @@
...
+async function getMany(userId, category, from, to) {
+  const queryArray = [];
+  const parametersArray = [...arguments].filter(parameter => parameter).length
+    ? [...arguments]
+    : [];
+
+  queryArray.push(userId ? 'user_id=$1' : '');
+  queryArray.push(category ? 'category=$2' : '');
+  queryArray.push(from ? 'spent_at>$3' : '');
+  queryArray.push(to ? 'spent_at<$4' : '');
```
dellachaise, line 34: What do you think, could you create some `enum` for this strings?

dellachaise (approved):
Great job 👍

src/services/expenses.js
```diff
@@ -0,0 +1,122 @@
...
+    title,
+    amount,
+    category,
+    note,
+  };
+}
+
+async function getMany(userId, category, from, to) {
+  const queryArray = [];
+  const parametersArray = [...arguments].filter(parameter => parameter).length
```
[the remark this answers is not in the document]
OlehMy: Yes, you're right) But, as i remember, `arguments` is object 🤔

src/services/expenses.js
OlehMy: Yes, when we use TS, I think it's a perfect case for use `enum` ))

OlehMy: Thanks a lot for your reviews 😉

PR #13, opened 2022-11-09
node_accounting-app-with-db

MaxManis:


sergii-nosachenko (changes requested):
To understand what your API service does I need to read your source code. For end-user you show nothing... How can somebody use your API?
Also the task says: `Implement CRUD page to manage categories` (but of course it's about expenses). 
Show me result of your work, present it. Show instructions with all possible end-points of your API or (better) create CRUD page(s) to manipulate the data.
Also check comments for previous task and fix similar mistakes here.

MaxManis:
CRUD page and other feathers implemented on other project but with a using of THIS server as a base;
[-> HERE](https://github.com/mate-academy/node_auth-app/pull/4)

PR #17, opened 2022-11-28
accounting app with data base implemented

artem-musii:


src/utils/setup.js
```diff
@@ -0,0 +1,10 @@
+/* eslint-disable no-console */
+'use strict';
+
+const { User } = require('../models/User');
+const { Expense } = require('../models/Expense');
+
+(async() => {
+  await User.sync({ force: true });
+  await Expense.sync({ force: true });
```
TheVovchik, line 9: do you need rewrite Data every time?

src/utils/models/User.js
```diff
@@ -0,0 +1,25 @@
...
+    allowNull: false,
+  },
+  name: {
+    type: DataTypes.STRING,
+    allowNull: false,
+  },
+}, {
+  tableName: 'users',
+  updatedAt: false,
+  createdAt: false,
```
TheVovchik, line 20: I also mark this to false, but mentors mark that better leave this columns to be created

src/utils/models/User.js
```diff
@@ -0,0 +1,25 @@
+'use strict';
+
+const { DataTypes } = require('sequelize');
+const { sequelize } = require('../../utils/db');
+
+const User = sequelize.define('User', {
+  id: {
+    type: DataTypes.INTEGER,
+    primaryKey: true,
+    autoIncrement: true,
```
TheVovchik, line 10: I think this one is true by default if you mark primaryKey on true

src/utils/models/Expense.js
```diff
@@ -0,0 +1,53 @@
...
+
+}, {
+  tableName: 'expenses',
+  updatedAt: false,
+});
+
+Expense.belongsTo(User, {
+  foreignKey: 'userId',
+  constraints: false,
+});
```
TheVovchik, line 49: interesting 👍

src/utils/models/Expense.js
```diff
@@ -0,0 +1,53 @@
...
+  note: {
+    type: DataTypes.STRING,
+    allowNull: false,
+  },
+  createdAt: {
+    type: DataTypes.DATE,
+    field: 'spentAt',
+    allowNull: false,
+    defaultValue: DataTypes.NOW,
+  },
```
TheVovchik, line 39: I think spentAt and createdAt are different. because we can spent one hour ago but create an expense now

src/utils/models/Expense.js
```diff
@@ -0,0 +1,53 @@
...
+  userId: {
+    type: DataTypes.INTEGER,
+    allowNull: false,
+  },
+  title: {
+    type: DataTypes.STRING,
+    allowNull: false,
+  },
+  amount: {
+    type: DataTypes.INTEGER,
```
TheVovchik, line 23: I use real data type to use decimal

src/utils/models/Expense.js
```diff
@@ -0,0 +1,53 @@
...
+    allowNull: false,
+  },
+  category: {
+    type: DataTypes.STRING,
+    allowNull: false,
+  },
+  note: {
+    type: DataTypes.STRING,
+    allowNull: false,
+  },
```
TheVovchik, line 33: I think note can be null, because user can write some note or not

src/services/users.js
```diff
@@ -0,0 +1,39 @@
...
+  });
+};
+
+const changeUser = async(userId, name) => {
+  const foundedUser = await User.findByPk(userId);
+
+  await foundedUser.update({ name });
+
+  return foundedUser.save();
+};
```
TheVovchik, line 31: so u find user in controller and here.
can we use just update and avoid 4 time call our database?
like we check if userId is correct === number, after we update direct by id(1 SQL call) and get updated or null

src/services/expenses.js
```diff
@@ -0,0 +1,72 @@
...
+  });
+};
+
+const changeExpense = async(expenseId, expenseData) => {
+  const foundedExpense = await Expense.findByPk(expenseId);
+
+  await foundedExpense.update(expenseData);
+
+  return foundedExpense.save();
+};
```
TheVovchik, line 63: same as with user update. avoid 4 call and do it in 2 SQL calls

src/services/expenses.js
```diff
@@ -0,0 +1,72 @@
...
+      if (from && new Date(expense.spentAt) < new Date(from)) {
+        return;
+      }
+
+      if (to && new Date(expense.spentAt) > new Date(to)) {
+        return;
+      }
+
+      return true;
+    });
```
TheVovchik, line 34: strange that u don't use SQL filters. I think u need to rewrite this one on sequelize SQL calls .findAll with clauses

src/createServer.js
```diff
@@ -0,0 +1,24 @@
...
+
+const { usersRouter } = require('./routes/users');
+const { expensesRouter } = require('./routes/expense');
+const app = express();
+
+function createServer() {
+  app.use(express.json());
+  app.use('/users', usersRouter);
+  app.use('/expenses', expensesRouter);
+}
```
TheVovchik, line 13: I am not sure but I think u need use CORS()

TheVovchik (approved):
leave some comments

gj 🔥

PR #27, opened 2023-01-24
Solution

andreyolin:


src/controllers/expenses.js
```diff
@@ -0,0 +1,77 @@
...
+  await expensesService.addOne(newExpense);
+
+  res.statusCode = 201;
+  res.send(newExpense);
+};
+
+const remove = (req, res) => {
+  const { expenseId } = req.params;
+
+  expensesService.deleteOne(expenseId);
```
iryna-lelyk, line 44: Add checking if expense is exist and only then delete it. You can call your function `expensesService.getById` to check it. And if there is no such expense, send status code 404

iryna-lelyk (approved):
Great job!

src/utils/db.js
```diff
@@ -0,0 +1,10 @@
+'use strict';
+
+const { Sequelize } = require('sequelize');
+
+const sequelize = new Sequelize('postgres', 'postgres', 'Dbpefkb7nbrf', {
+  host: 'localhost',
+  dialect: 'postgres',
+});
```
Olexandr6, line 8: Would be good to move DB configuration to `.env` file

Olexandr6 (approved):
Good job, Andriy 👍

PR #47, opened 2023-02-27
solution

Gintako666:


src/services/users.js
```diff
@@ -0,0 +1,43 @@
+'use strict';
+
+const { User } = require('../models/User');
+
+async function getAll() {
+  const users = await User.findAll({
+    order: [
+      'id',
+    ],
```
DarkMistyRoom, line 9: if id is autoincremented, ok. but I'd preferred to sort by `createdAt`

DarkMistyRoom (changes requested):
Good job. Let's handle errors

PR #49, opened 2023-03-11
db added

svitjojo:


src/services/expenses.js
```diff
@@ -0,0 +1,75 @@
...
+          };
+
+          delete select.to;
+          break;
+
+        default:
+          break;
+      }
+    });
+  }
```
nzrii-kysel, line 41: you could just use 2 ifs instead of this forEach

```suggestion
function getAllByQuery(query) {
  const select = { ...query };
  
  if (select.hasOwnProperty('from')) {
    select.spentAt = {
      [Op.gte]: dateNormalize(select.from),
    };
    delete select.from;
  }

  if (select.hasOwnProperty('to')) {
    select.spentAt = {
      ...select.spentAt,
      [Op.lte]: dateNormalize(select.to),
    };
    delete select.to;
  }
```

src/utils/dateNormalize.js
```diff
@@ -0,0 +1,15 @@
...
+  const dateParts = dateString.split(' ');
+
+  const date = dateParts[0];
+  const time = dateParts[1];
+  const timezone = dateParts[2];
+
+  const normalized = `${date} ${time}+${timezone}`;
+
+  return normalized;
+};
```
nzrii-kysel, line 13: I can't think of any reason why you need this function, I believe it only hurts the functionality

nzrii-kysel (approved):
well done

src/utils/dateNormalize.js
svitjojo: Ohh, I spent hours fixing the problem with date selection, because decided to make spentAt property automatically created by DB, but a problem is that it has its own date format, and js can't recognize it, it throws an error and crashes the server. So the only way I found is this function.

src/services/expenses.js
svitjojo: You are right

src/utils/dateNormalize.js
nzrii-kysel: sounds like a `kostyl'` then

svitjojo: Maybe there is a better way, but I didn't find it.

src/db/db.js
```diff
@@ -0,0 +1,14 @@
+'use strict';
+
+const { Sequelize } = require('sequelize');
+
+require('dotenv').config();
+
+const password = process.env.DB_PASSWORD;
+
+const sequelize = new Sequelize('postgres', 'postgres', password, {
+  host: 'localhost',
```
anna-inozemets, line 10: We also can use reject for unauthorized users

src/routes/expenses.js
```diff
@@ -0,0 +1,18 @@
+'use strict';
+
+const express = require('express');
+const expenseController = require('../controllers/expenses.js');
+
+const router = express.Router();
+
+router.get('/', expenseController.getAll);
+
```
anna-inozemets, line 9: I guessed it is unnecessary empty line, and others like this

src/services/expenses.js
```diff
@@ -0,0 +1,60 @@
+'use strict';
+
+const { Expense } = require('../models/Expense.js');
+const { Op } = require('sequelize');
+
+async function getAll() {
```
anna-inozemets, line 6: If we use async, we need use await, right?

src/services/expenses.js
```diff
@@ -0,0 +1,60 @@
...
+  return Expense.create(body);
+}
+
+async function remove(id) {
+  await Expense.destroy({
+    where: { id },
+  });
+}
+
+function update(id, body) {
```
anna-inozemets, line 47: I guess, that getById, create, remove and this one should also be async

src/setup.js
```diff
@@ -0,0 +1,7 @@
+'use strict';
+
+const { User } = require('./models/User');
+const { Expense } = require('./models/Expense');
+
+User.sync({ force: true });
```
anna-inozemets, line 6: Maybe alert here is better instead of force

anna-inozemets (approved):
Really good code!)

src/services/expenses.js
svitjojo: Hmm, it returns the promise by itself, so I don't think we need async await here.

.gitignore
```diff
@@ -2,6 +2,9 @@
 .idea
 .vscode
 
+# PRIVATE
```
ArturErmolenko, line 5: checklist 2

ArturErmolenko (approved):
GJ 👍

PR #57, opened 2023-03-25
solution

Vladyslav317:


src/utils/db.js
```diff
@@ -0,0 +1,10 @@
+'use strict';
+
+const { Sequelize } = require('sequelize');
+
+const sequelize = new Sequelize('postgres', 'postgres', 'postgres', {
+  host: 'localhost',
+  dialect: 'postgres',
+});
```
OlEzhA0, line 8: in real project you need to keep this data in `.env` file

src/components/users/users.service.js
```diff
@@ -0,0 +1,55 @@
+'use strict';
+
+const usersRepository = require('./users.repository');
+
+function create(name) {
+  return usersRepository.create(name);
+}
+
+async function findAll() {
+  const [...users] = await usersRepository.findAll();
```
OlEzhA0, line 10: why do we need it?
`const users = ...`

OlEzhA0 (approved):
hey! gj
left few comments

src/components/users/users.service.js
Vladyslav317: I can check if users exist without using users.length

src/components/users/users.controller.js
```diff
@@ -0,0 +1,96 @@
...
+const remove = async(req, res) => {
+  const { id } = req.params;
+
+  const user = await usersService.findByUserId(id);
+
+  if (!user) {
+    res.sendStatus(404);
+
+    return;
+  }
```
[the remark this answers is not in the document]
Vladyslav317: I suppose it is. I should check if user exists then I can use his id to remove it from database.

PR #62, opened 2023-04-11
Solution

stanislavcodes:
<details>
<summary>Local Tests with PostgreSQL</summary>
<br/>

image
</details>


src/controllers/expenses.js
```diff
@@ -0,0 +1,119 @@
...
+  const { expenseId } = req.params;
+  const foundExpense = await expensesService.getById(expenseId);
+
+  if (!foundExpense) {
+    res.sendStatus(404);
+
+    return;
+  }
+
+  await expensesService.update(expenseId, { ...req.body });
```
rashyd-hasratov, line 106: It's better to take just those properties from `req.body` that we need, not just to restructure the whole object

src/createServer.js
```diff
@@ -0,0 +1,20 @@
...
+const cors = require('cors');
+const { router: usersRouter } = require('./routes/users.js');
+const { router: expensesRouter } = require('./routes/expenses.js');
+
+function createServer() {
+  const app = express();
+
+  app.use(cors());
+  app.use('/users', express.json(), usersRouter);
+  app.use('/expenses', express.json(), expensesRouter);
```
rashyd-hasratov, line 13: It would be better to write this part in such a way

```suggestion
  const app = express();

  app.use(cors());
  app.use(express.json());
  app.use('/users', usersRouter);
  app.use('/expenses', expensesRouter);
```

rashyd-hasratov (approved):
Great work! Just a couple of suggestions that may help you avoid a reject

Oleksandr-kopaevich (approved):
Hey @stanislavcodes, good job with the task!

PR #64, opened 2023-04-27
solution

eugenekh81:
Done with Sequelize

Please review ) TY

### DIFF
--- a/src/controllers/users.js
+++ b/src/controllers/users.js
@@ -1,0 +1,57 @@
+'use strict';
+
+const userServices = require('../services/users');
+
+const getAll = async(req, res) => {
+  const users = await userServices.getAll();
+
+  res.send(users);
+};
+
+const getOne = async(req, res) => {
+  const { id } = req.params;
+  const user = await userServices.getOne(id);
+
+  res.send(user);
+};
+
+const add = async(req, res) => {
+  const { name } = req.body;
+
+  if (typeof name !== 'string') {
+    res.sendStatus(422);
+
+    return;
+  }
+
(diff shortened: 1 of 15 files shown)


src/controllers/users.js
```diff
@@ -0,0 +1,57 @@
+'use strict';
+
+const userServices = require('../services/users');
+
+const getAll = async(req, res) => {
+  const users = await userServices.getAll();
```
DarkMistyRoom, line 6: wrapp all async operations in try catch(or use then catch) to avoid uncatched errors

src/controllers/users.js
```diff
@@ -0,0 +1,57 @@
...
+
+  res.send(users);
+};
+
+const getOne = async(req, res) => {
+  const { id } = req.params;
+  const user = await userServices.getOne(id);
```
DarkMistyRoom, line 13: check if `id` was passed and if `user` was found

DarkMistyRoom: apply analogic checks to all other controllers

src/services/users.js
```diff
@@ -0,0 +1,48 @@
+'use strict';
+
+const User = require('../models/users');
+
```
DarkMistyRoom, line 4: use normalization function. There are such fields as `createdAt`, `updatedAt `which is no need to send

DarkMistyRoom: same for expenses

DarkMistyRoom (changes requested):
Good job! But let's handle errors

nazarmatsevych (approved):
👍

PR #66, opened 2023-05-01
solution

Zelinskyi-Serhii:


src/utils/db.js
```diff
@@ -0,0 +1,10 @@
+'use strict';
+
+const { Sequelize } = require('sequelize');
+
+const sequelize = new Sequelize('postgres', 'postgres', '1111', {
```
juljaVysotska, line 5: Consider using environment variables instead of hardcoded values for database credentials and host to increase security and flexibility.

src/services/expenses.js
```diff
@@ -0,0 +1,91 @@
...
+  category,
+  note = 'Any notes',
+}) => {
+  return Expense.create({
+    userId, spentAt, title, amount, category, note,
+  });
+};
+
+const removeExpense = (expenseId) => {
+  Expense.destroy({
```
juljaVysotska, line 54: You should return the result of the Expense.destroy() call.

juljaVysotska (approved):
Good job!🔥

PR #67, opened 2023-05-04
accounting-app-with-db was implement

KuharchukDmytro:


vladyslav-mosor (approved):

src/utils/db.js
```diff
@@ -0,0 +1,19 @@
+'use strict';
+
+const { Sequelize } = require('sequelize');
+
+require('dotenv').config();
```
And678, line 5: Good idea to .env for config

src/services/usersService.js
```diff
@@ -0,0 +1,73 @@
...
+};
+
+const reset = async() => {
+  const users = await getAll();
+
+  return sequelize.transaction(async(t) => {
+    for (const user of users) {
+      await removeUser(user.id);
+    }
+  });
```
And678, line 63: This can be improved, using transaction is good, but not needed as SQL databases have `TRUNCATE` command that removed all data from table in only one request. You can do it by `UserModel.destroy({truncate: true})` or `UserModel.truncate()`.
[Docs1](https://sequelize.org/api/v6/class/src/model.js~model#static-method-truncate), [docs2](https://sequelize.org/api/v6/class/src/model.js~model#static-method-destroy)

.env
And678, line 1: While its ok for this task, remember to **NEVER** push any keys/passwords/connectionStrings to code repo for a real project.

src/services/usersService.js
```diff
@@ -0,0 +1,73 @@
...
+const updateUser = async(userId, name) => {
+  if (!getById(userId)) {
+    return null;
+  }
+
+  const updatedUser = await UserModel.update({ name }, {
+    where: {
+      id: userId,
+    },
+  });
```
And678, line 51: Model.update doesn't return the row instance. Check the [doc](https://sequelize.org/api/v6/class/src/model.js~model#static-method-update).

What you need is to take the instance from getById, and call [update](https://sequelize.org/api/v6/class/src/model.js~model#instance-method-update) on it, not the static model method

And678 (changes requested):
Hello! You need add a way to start your app. For example an `app.js` which calls `createServer().listen`
I tried it, your app doesn't launch.
image

Please test your work before submitting.
If you have trouble launching database locally, reach out to QnA.
For example you can start a new database instantly using [docker](https://www.docker.com/) like this: `docker run --name mate-postgres -p 5432:5432 -e POSTGRES_PASSWORD=12345678 -d postgres`

src/createServer.js
```diff
@@ -0,0 +1,24 @@
...
+
+function createServer() {
+  const cors = require('cors');
+  const { router: usersRouter } = require('./routes/users.js');
+  const { router: expensesRouter } = require('./routes/expenses.js');
+  const expensesService = require('./services/expensesService.js');
+  const userService = require('./services/usersService.js');
+
+  expensesService.reset();
+  userService.reset();
```
And678, line 13: No need to reset, actually, we have a database here to keep the data, even if the app has stopped
```suggestion
```

PR #68, opened 2023-05-04
accounting app with db done

rashyd-hasratov:


src/index.js
```diff
@@ -1 +1,8 @@
 'use strict';
+
+const { createServer } = require('./createServer');
+
+createServer()
+  .listen(3000, () => {
+    global.console.log('Server is running on 3000 port');
```
juljaVysotska, line 7: Instead of using 'global.console.log', it's more common to use 'console.log' directly.

src/utils/db.js
```diff
@@ -0,0 +1,10 @@
+'use strict';
+
+const { Sequelize } = require('sequelize');
+
+const sequelize = new Sequelize('matepostgres', 'postgres', 'lek', {
```
juljaVysotska, line 5: Hardcoded database credentials can lead to security issues. Consider using environment variables to store sensitive data.

src/services/expenses.js
```diff
@@ -0,0 +1,82 @@
+'use strict';
+
+const { Expense } = require('../models/expenses');
+
+const getAll = async(filterOptions) => {
+  const expenses = await Expense.findAll();
```
juljaVysotska, line 6: Apply the filterOptions directly in the Expense.findAll() to reduce the number of fetched expenses.

src/setup.js
```diff
@@ -0,0 +1,7 @@
+'use strict';
+
+const { User } = require('./models/users');
+const { Expense } = require('./models/expenses');
+
+User.sync({ force: true });
+Expense.sync({ force: true });
```
juljaVysotska, line 7: Using 'force: true' in the sync method will cause the table to be dropped and recreated every time the application is started. This will result in a loss of data. Consider removing the 'force' option or setting it to false.

juljaVysotska (changes requested):

PR #70, opened 2023-05-07
accounting app with db implemented

vladyslav-mosor:


src/database/sequelize.js
```diff
@@ -0,0 +1,8 @@
+'use strict';
+
+const Sequelize = require('sequelize');
+
+module.exports = new Sequelize('postgres', 'postgres', '8888', {
```
juljaVysotska, line 5: The password for the PostgreSQL connection should not be hardcoded. Use environment variables instead.

src/database/sequelize.js
```diff
@@ -0,0 +1,8 @@
...
+  host: 'localhost',
```
juljaVysotska, line 6: The database host should not be hardcoded. Use environment variables instead.

src/controllers/expenses.js
```diff
@@ -0,0 +1,77 @@
+'use strict';
+
+const expensesService = require('../services/expenses.js');
+const usersService = require('../services/users.js');
+
+const getAll = async(req, res) => {
```
juljaVysotska, line 6: Use destructuring assignment to obtain 'query' from 'req'. Example: const { query } = req;

juljaVysotska (approved):
Good job!

PR #73, opened 2023-05-14
Solution

MykhailoIvchenko:


src/services/users.js
```diff
@@ -0,0 +1,72 @@
+'use strict';
+
+const { DataTypes } = require('sequelize');
+const { sequelize } = require('../db');
+const { Expense } = require('../services/expenses');
+
+function normalize({ id, name }) {
```
nazarmatsevych, line 7: Function 'normalize' is not being used in this file. Consider removing it if it's not needed.

src/services/expenses.js
```diff
@@ -0,0 +1,151 @@
...
+});
+
+function normalize({ id, userId, title, spentAt, amount, category, note }) {
+  return {
+    id, userId, title, spentAt, amount, category, note,
+  };
+}
+
+function getAll() {
+  try {
```
nazarmatsevych, line 42: It's better to use 'async' and 'await' for error handling in asynchronous functions.

PR #91, opened 2023-06-17
add postgres db to accounting app

TarasHoliuk:
Added Postgres db to the accounting app


src/services/expenses/expenses.js
```diff
@@ -0,0 +1,99 @@
...
+    WHERE id = ${id}
+    RETURNING id
+  `;
+
+  return response[0];
+};
+
+const update = async (id, updatedExpense) => {
+  const { title, amount, category, note, spentAt } = updatedExpense;
+
```
TarasHoliuk, line 76: there should be smth like that:
```
  const setProperties = Object.entries(updatedExpense)
    .filter(([_key, value]) => value !== undefined)
    .map(([key, value]) => `${key} = ${value}`)
    .join(', ');
  ```
  
 but it doesn't work due to the way `postgres` library parses the template (!?)
 even if 
 `setProperties === 'title = ${title},
      amount = ${amount},
      category = ${category},
      note = ${note},
      spentAt = ${spentAt}'`
      
   the same with filter options in `getAll` function (now it works if all params are defined)
   in the real project, I'd use smth like Sequelize or Prisma library but to get used to postgress query syntax and not make task solution more complex I leave `postgres.Sql`

File: .env [hunk 2]
```diff
@@ -0,0 +1,6 @@
+PORT = 5000
```
YuriiHoliuk, line 1: You shouldn't commit .env, only .env.sample. Let's do it like in real project.

src/createServer.js
```diff
@@ -0,0 +1,29 @@
...
+
+const cors = require('cors');
+
+const express = require('express');
+
+function createServer() {
+  const app = express();
+
+  usersService.resetStore();
+  expensesService.resetStore();
```
YuriiHoliuk, line 16: Why do we need it?

src/services/expenses/expenses.js
```diff
@@ -0,0 +1,99 @@
+'use strict';
+
+const { v4: uuid } = require('uuid');
+
+const { db } = require('../../db');
+
+const { validateExpenseData } = require('./validation');
+
+const resetStore = async () => {
```
YuriiHoliuk, line 9: Ok, got it. But better to use the migrations concept.

src/services/expenses/expenses.js
```diff
@@ -0,0 +1,99 @@
...
+const getAll = async ({ userId, categories, from, to }) => {
+  return db`
+    SELECT *
+    FROM expenses
+    WHERE
+      userId = ${userId}
+      AND
+      spentAt BETWEEN ${from} AND ${to}
+      AND
+      category IN (${categories.join(', ')})
```
YuriiHoliuk, line 32: It will work as long as you and 5 friends will use this app.
Otherwise, you need DB indices.

src/services/expenses/expenses.js
```diff
@@ -0,0 +1,99 @@
...
+    userId UUID,
+    title TEXT,
+    category TEXT,
+    amount NUMERIC,
+    spentAt TIMESTAMP,
+    note TEXT
+  )`;
+};
+
+const getAll = async ({ userId, categories, from, to }) => {
```
YuriiHoliuk, line 23: It's a common design mistake when the Service is tightly coupled with the DB.
Better to introduce a Repository concept, ideally with dependency injection.

src/services/expenses/expenses.js
```diff
@@ -0,0 +1,99 @@
...
+
+const { db } = require('../../db');
+
+const { validateExpenseData } = require('./validation');
+
+const resetStore = async () => {
+  await db`DROP TABLE IF EXISTS expenses`;
+
+  await db`CREATE TABLE expenses (
+    id UUID PRIMARY KEY,
```
YuriiHoliuk, line 13: Better to use sequence id. Watch Turskyi😅

src/services/expenses/expenses.js
YuriiHoliuk: It's ok to have every field manually. Explicit over implicit.

src/sharedValidation.js
YuriiHoliuk, line 1: src/shared/validation.js

src/db.js
```diff
@@ -0,0 +1,15 @@
...
+
+const sql = postgres({
+  host: process.env.POSTGRES_HOST,
+  port: process.env.POSTGRES_PORT,
+  user: process.env.POSTGRES_USER,
+  password: process.env.POSTGRES_PASSWORD,
+  database: 'accounting_app_db',
+});
+
+module.exports = { db: sql };
```
YuriiHoliuk, line 15: Raw SQL is impossible to support. Better to use some ORM or query library.
I highly recommend trying Prisma. TypeORM is also an option.

YuriiHoliuk (changes requested):
Looks good in general.
In addition to comments in PR, you can add tests and GH action to run them, Dockerfile and docker-compose to run it with a single command.
If you want to make it production-ready and learn more, ping me. If not, I will approve it next time🙂.

.env
TarasHoliuk: yeah, TU :+1:

src/createServer.js
TarasHoliuk: actually, we don't need
it was needed in the previous task to pass tests

I'll get rid of it

src/services/expenses/expenses.js
TarasHoliuk: Not sure how it should look like
If the table with such a name exists we can check if the schema is correct and if the table is not empty - migrate data to fix the schema if it's needed.
But it's the very first version yet, and there can't exist an "outdated" version

src/storages/expenses.ts
```diff
@@ -0,0 +1,29 @@
+import { PrismaClient } from '@prisma/client'
+
+export default class ExpensesStorage {
```
TarasHoliuk, line 3: I was trying to create a reusable class with the correct type of prismaClient.model by passing the key in the constructor but it seems like that is not so easy :)
e.g.
```
const expensesStorage = new Storage(prismaClient, 'expenses');

expensesStorage.findMany() // returns expenses
```

TarasHoliuk:
It's still not perfect for sure, but a bit better I guess :)

src/controllers/expenses.ts
```diff
@@ -0,0 +1,82 @@
...
+
+    const expenseData = { userId, spentAt, title, amount, category, note }
+
+    try {
+      const expense = await this.expensesService.create(expenseData)
+
+      res.status(201).send(expense)
+    } catch (error) {
+      const message =
+        error instanceof Error ? error.message : 'Error creating expense'
```
YuriiHoliuk, line 51:
```suggestion
      const message = error instanceof Error
        ? error.message
        : 'Error creating expense'
```

src/controllers/expenses.ts
YuriiHoliuk, line 1: Шо ти сильно модний і капки з комою не пишеш?

src/controllers/users.ts
```diff
@@ -0,0 +1,77 @@
...
+import { Request, Response } from 'express'
+import { User } from '@prisma/client'
+import type { UserResponse } from '../types'
+
+export default class UsersController {
+  private usersService: UsersService;
+
+  constructor (service: UsersService) {
+    this.usersService = service
+  }
```
YuriiHoliuk, line 13:
```suggestion
  constructor (private usersService: UsersService) {
  }
```

src/routes/expenses.ts
```diff
@@ -0,0 +1,31 @@
+'use strict'
+
+import { Router } from 'express'
+import { ExpensesController } from '../controllers'
+import { ExpensesService } from '../services'
+
+export default class ExpensesRouter {
```
YuriiHoliuk, line 7: I would never use default export

src/services/expenses.ts
```diff
@@ -0,0 +1,113 @@
...
+import type {
+  RawExpensesFilter,
+  ExpensesFilter,
+  ExpensesInput
+} from '../types'
+
+export default class ExpensesService {
+  private constructor (
+    private storage: ExpensesStorage,
+    private serviceRegistry: ServiceRegistry
```
YuriiHoliuk, line 16: Ideally, a service shouldn't know about the existence of a registry.

src/services/expenses.ts
```diff
@@ -0,0 +1,113 @@
...
+        amount: true,
+        category: true,
+        note: true
+      },
+      where: {
+        userId: { equals: filter.userId },
+        spentAt: { gte: filter.from, lte: filter.to },
+        category: { in: filter.categories }
+      }
+    })
```
YuriiHoliuk, line 46: Looks like storage exposes too many low-level details.

src/services/expenses.ts
```diff
@@ -0,0 +1,113 @@
+'use strict'
```
YuriiHoliuk, line 1: It's not needed with the typescript.

src/services/serviceRegistry.ts
```diff
@@ -0,0 +1,34 @@
+import UsersService from './users'
+import ExpensesService from './expenses'
```
YuriiHoliuk, line 2: You have circular dependencies. Services depend on registry and vice versa.

src/services/users.ts
```diff
@@ -0,0 +1,37 @@
...
+
+import { UsersStorage } from '../storages'
+import { ServiceRegistry } from './serviceRegistry'
+
+export default class UsersService {
+  private constructor (
+    private storage: UsersStorage,
+    private serviceRegistry: ServiceRegistry
+  ) {
+    this.serviceRegistry.register('UsersService', this)
```
YuriiHoliuk, line 11: What if you copy this class creating another service and will forget to change this string?

src/storages/expenses.ts
YuriiHoliuk: Prisma exports a lot of types. Probably you might use them.

But I see storage-like classes with a bit higher level of abstraction. They should call prisma methods and pass arguments to them. findAll can exist but methods like findByEmail also have a place to be.
Or you might not use this abstraction at all and use prismaClient in services.
But approach with services is better.

YuriiHoliuk (approved):
🔥

Dockerfile
```diff
@@ -0,0 +1,26 @@
+FROM node:alpine
+
+WORKDIR /app
+
+# COPY package.json and package-lock.json files
+COPY package*.json ./
+
+# generated prisma files
+COPY prisma ./prisma/
```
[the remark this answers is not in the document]
TarasHoliuk: Indeed, don't know why have I committed them

prisma/migrations/20230816145808_initialize/migration.sql
```diff
@@ -0,0 +1,36 @@
+-- CreateTable
+CREATE TABLE "User" (
+    "id" SERIAL NOT NULL,
+    "createdAt" TIMESTAMP(3) NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    "updatedAt" TIMESTAMP(3) NOT NULL,
```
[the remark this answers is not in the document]
TarasHoliuk: :thinking: :+1:

src/createServer.ts
```diff
@@ -0,0 +1,48 @@
...
+    expenses: ExpensesService.create(expensesStorage, serviceRegistry)
+  }
+}
+
+const createRouters = (services: ReturnType<typeof createServices>) => {
+  return {
+    users: UsersRouter.create(services.users),
+    expenses: ExpensesRouter.create(services.expenses)
+  }
+}
```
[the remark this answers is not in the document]
TarasHoliuk: wow, didn't know it. But probably you can find lib for almost anything

:smile:

src/controllers/expenses.ts
TarasHoliuk: Мабуть дефолтні налаштування якогось конфігу - я додавав)

src/controllers/users.ts
TarasHoliuk: indeed - TU :+1:

src/services/expenses.ts
TarasHoliuk: hm, why? The idea is to use it to get another service's methods. How can we do it w/o having a serviceRegistry as one of the service's properties?

Or do you mean a particular service shouldn't know (in case we have some abstract BaseService class just pass `serviceRegistry` as one of `super` properties, etc?)

src/services/expenses.ts
TarasHoliuk: Got it

I guess we need some type for a filter of each entity or even better, a general filter parser to more low-level things (or both)

src/services/users.ts
TarasHoliuk: the ServiceRegistry will throw the error: `ServiceRegistry already contains ${name}`. But yes, better to do it automatically

PR #94, opened 2023-07-14
Develop

maksym-nemera:


src/createServer.js
```diff
@@ -0,0 +1,18 @@
...
+const createServer = () => {
+  const app = express();
+
+  app.use('/users', express.json(), userRouter);
+  app.use('/expenses', express.json(), expenseRouter);
+
+  return app;
+};
+
+createServer();
```
33Orange, line 16: You create a server 2 times. in Main.js and in createServer.

src/controllers/users.js
```diff
@@ -0,0 +1,93 @@
...
+      ['createdAt', 'ASC'],
+    ],
+  });
+
+  res.json(users.map(normalize));
+};
+
+const getOneUser = async(req, res) => {
+  const { userId } = req.params;
+  const findUser = await User.findByPk(userId);
```
33Orange, line 24: Let's add some errors if we don't have the required data.
Same about user service

```suggestion
  const { userId } = req.params;
  
  if (!userdId) {
    some error
  }
  
  const findUser = await User.findByPk(userId);
```

src/controllers/users.js
```diff
@@ -0,0 +1,93 @@
...
+
+  if (!findUser) {
+    return res.sendStatus(404);
```
33Orange, line 27: Let's add some error messages

src/controllers/users.js
```diff
@@ -0,0 +1,93 @@
...
+  if (!name) {
+    res.status(400);
+    res.json({ error: 'Name is required' });
+
+    return;
+  }
+
+  const newUser = await User.create({
+    name,
+  });
```
33Orange, line 45: Same about service here and in all other cases

src/controllers/users.js
```diff
@@ -0,0 +1,93 @@
...
+
+  res.status(201);
+  res.json(normalize(newUser));
+};
+
+const removeUser = async(req, res) => {
+  const { userId } = req.params;
+
```
33Orange, line 53: Let's add some errors if we don't have the required data.
```suggestion
  const { userId } = req.params;
  if (!userId) {
    error
  }
```

src/controllers/users.js
```diff
@@ -0,0 +1,93 @@
...
+  await User.destroy({
+    where: { id: userId },
+  });
+
+  res.sendStatus(204);
+};
+
+const updateUser = async(req, res) => {
+  const { name } = req.body;
```
33Orange, line 68: Add some error if you don't have a `name` to update

src/controllers/expenses.js
```diff
@@ -0,0 +1,133 @@
...
+      [Op.between]: [fromDate, toDate],
+    };
+  }
+
+  const filteredExpenses = await Expense.findAll({
+    where: whereClause,
+    order: [
+      ['createdAt', 'ASC'],
+    ],
+  });
```
33Orange, line 45: Same about the new service layout. 
Create a separate service for expenses

src/controllers/expenses.js
```diff
@@ -0,0 +1,133 @@
...
+  res.json(filteredExpenses);
+};
+
+const getOneExpense = async(req, res) => {
+  const { expenseId } = req.params;
+
+  const findExpense = await Expense.findByPk(expenseId);
+
+  if (!findExpense) {
+    return res.sendStatus(404);
```
33Orange, line 56: Same about the error messages.

33Orange (changes requested):
Almost done

PR #96, opened 2023-07-25
made task node-accounting-app-with-db

vladyslav-lapshyn:


src/createServer.js
```diff
@@ -0,0 +1,20 @@
...
+function createServer() {
+  const app = express();
+
+  app.use('/users', express.json(), userRouter);
+  app.use('/expenses', express.json(), expensesRouter);
+
+  return app;
+}
+
+createServer();
```
33Orange, line 16: You create a server twice. Once in a `main.js` and once in a` createServer.js`

src/controllers/users.controller.js
```diff
@@ -0,0 +1,86 @@
...
+    users.map(usersService.normalize)
+  );
+};
+
+const getOne = async(req, res) => {
+  const { userId } = req.params;
+  const foundUser = await usersService.getById(userId);
+
+  if (!foundUser) {
+    res.sendStatus(404);
```
33Orange, line 18: Let's add some error messages to the errors, to describe a problem

src/controllers/users.controller.js
```diff
@@ -0,0 +1,86 @@
...
+  const users = await usersService.getAll();
+
+  res.send(
+    users.map(usersService.normalize)
+  );
+};
+
+const getOne = async(req, res) => {
+  const { userId } = req.params;
+  const foundUser = await usersService.getById(userId);
```
33Orange, line 15: Let's check the required data before using business logic
```suggestion
  const { userId } = req.params;
  
  if (!userdId) {
    some error
  } 
  
  const foundUser = await usersService.getById(userId);
```

src/controllers/users.controller.js
```diff
@@ -0,0 +1,86 @@
...
+  res.send(
+    usersService.normalize(foundUser)
+  );
+};
+
+const add = async(req, res) => {
+  const { name } = req.body;
+
+  if (!name) {
+    res.sendStatus(400);
```
33Orange, line 32: Same about the errors messages here and in all other cases

src/controllers/users.controller.js
```diff
@@ -0,0 +1,86 @@
...
+  }
+
+  const newUser = await usersService.create(name);
+
+  res.statusCode = 201;
+  res.send(usersService.normalize(newUser));
+};
+
+const remove = async(req, res) => {
+  const { userId } = req.params;
```
33Orange, line 44: Same about checking required data here and in all other cases

src/controllers/users.controller.js
```diff
@@ -0,0 +1,86 @@
...
+
+  if (!foundUser) {
+    res.sendStatus(404);
+
+    return;
+  }
+
+  const { name } = req.body;
+
+  if (typeof name !== 'string') {
```
33Orange, line 69: In this case, it can be an empty string

src/controllers/expenses.controller.js
```diff
@@ -0,0 +1,105 @@
...
+    categories,
+  });
+
+  res.send(
+    expenses.map(expensesService.normalize)
+  );
+};
+
+const getOne = async(req, res) => {
+  const { expenseId } = req.params;
```
33Orange, line 27: SAme about the required data here and in all other cases in an expenses controller

33Orange (changes requested):
Almost done

PR #98, opened 2023-07-26
solution

okosohor:


src/controllers/users.js
```diff
@@ -0,0 +1,80 @@
...
+  res.status(201);
+  res.send(newUser);
+};
+
+const getOneUser = async(req, res) => {
+  const { userId } = req.params;
+  const foundUser = await usersService.getUserByID(userId);
+
+  if (!foundUser) {
+    res.sendStatus(404);
```
33Orange, line 31: Same here and in all other cases

src/controllers/users.js
```diff
@@ -0,0 +1,80 @@
...
+  }
+
+  const newUser = await usersService.createUser(name);
+
+  res.status(201);
+  res.send(newUser);
+};
+
+const getOneUser = async(req, res) => {
+  const { userId } = req.params;
```
33Orange, line 27: Let's check required data before using business logic

```suggestion
  const { userId } = req.params;
  if (!userId) {
    some error
  }
```

src/controllers/users.js
```diff
@@ -0,0 +1,80 @@
...
+  if (!foundUser) {
+    res.sendStatus(404);
+
+    return;
+  }
+
+  usersService.updateUser({
+    userId,
+    name,
+  });
```
33Orange, line 69: What would be if we don't have a name?
Let's add some check for required data

33Orange (changes requested):
Almost done

PR #99, opened 2023-07-27
Develop

growingananas:


src/controllers/userController.js
```diff
@@ -0,0 +1,91 @@
...
+
+const userServices = require('../services/userServices');
+
+const getAll = async(req, res) => {
+  try {
+    const users = await userServices.getAll();
+
+    res.send(users);
+  } catch (error) {
+    res.sendStatus(400);
```
33Orange, line 11: Let's add some messages to the errors to describe a problem.

src/controllers/userController.js
```diff
@@ -0,0 +1,91 @@
...
+  res.statusCode = 201;
+  res.send(newUser);
+};
+
+const getOne = async(req, res) => {
+  const { id } = req.params;
+  const foundUser = await userServices.getUserById(id);
+
+  if (!foundUser) {
+    res.sendStatus(404);
```
33Orange, line 35: Same here and in all other cases

src/controllers/userController.js
```diff
@@ -0,0 +1,91 @@
...
+
+    return;
+  }
+
+  try {
+    await userServices.remove(id);
+    res.status(204).send({ message: 'User has been deleted successfully' });
+  } catch (error) {
+    res.status(500).json({ message: 'Error occurred while removing the user' });
+  }
```
33Orange, line 59: Move this `try catch` to the service. 

```suggestion
    await userServices.remove(id);
```

src/controllers/expensesController.js
```diff
@@ -0,0 +1,122 @@
...
+const expenseServices = require('../services/expenseServices');
+const userServices = require('../services/userServices');
+
+const getAll = async(req, res) => {
+  try {
+    const expenses = await expenseServices.getAllExpenses(req.query);
+
+    res.status(200).send(expenses);
+  } catch (error) {
+    res.sendStatus(400);
```
33Orange, line 12: Same about the error messages in the expenses

src/services/expenseServices.js
```diff
@@ -0,0 +1,102 @@
+'use strict';
+
+const Expense = require('../models/expense');
+
+const validateData = (request) => {
```
33Orange, line 5: Let's move it to the separate file `helpers`

33Orange (changes requested):
Almost done

PR #100, opened 2023-07-27
added expenses, users routers, controllers, services

AlexKherson:


src/controllers/users.controllers.js
```diff
@@ -0,0 +1,73 @@
+'use strict';
+
+const { ApiError } = require('../errorHandler/errorHandler');
+const { UserService } = require('../services/users.service');
+
+const getAllUsersController = async(req, res) => {
+  const userService = new UserService();
```
33Orange, line 7: Use [a singleton pattern](https://www.freecodecamp.org/news/singleton-design-pattern-with-javascript/) if you want to work with classes.
Create `UserService` once instead of re-creating it every time

src/controllers/users.controllers.js
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  res.sendStatus(204);
+};
+
+const updateUsersController = async(req, res, next) => {
+  const usersService = new UserService();
+  const { userId } = req.params;
+  const newUser = req.body;
+
```
33Orange, line 57: Let's add some checks for existing data before we execute service methods

src/controllers/expenses.controller.js
```diff
@@ -0,0 +1,80 @@
+'use strict';
+
+const { ApiError } = require('../errorHandler/errorHandler');
+const { ExpensesService } = require('../services/expenses.service');
+
+const getAllExpensesController = async(req, res, next) => {
+  const expensesService = new ExpensesService();
```
33Orange, line 7: Same about a singleton class

src/controllers/expenses.controller.js
```diff
@@ -0,0 +1,80 @@
...
+  }
+
+  res.sendStatus(204);
+};
+
+const updateExpenseController = async(req, res, next) => {
+  const expensesService = new ExpensesService();
+  const { expenseId } = req.params;
+  const newExpense = req.body;
+
```
33Orange, line 64: Same about the checks

33Orange (changes requested):
Almost done

PR #101, opened 2023-07-28
Solution

julshvets:


src/controllers/users.controllers.js
```diff
@@ -0,0 +1,90 @@
...
+    normalizeUser(user)
+  )));
+};
+
+const getUser = async(request, response) => {
+  const { userId } = request.params;
+  const foundUser = await usersServices.getUserById(userId);
+
+  if (!foundUser) {
+    response.sendStatus(404);
```
33Orange, line 19: Let's add some messages to describe the error.

src/controllers/users.controllers.js
```diff
@@ -0,0 +1,90 @@
...
+  }
+
+  response.send(normalizeUser(foundUser));
+};
+
+const createUser = async(request, response) => {
+  const { name } = request.body;
+
+  if (!name) {
+    response.sendStatus(400);
```
33Orange, line 31: Same here and in all other cases

src/controllers/expenses.controllers.js
```diff
@@ -0,0 +1,105 @@
...
+  const expenses = await expensesServices.getFiltered(request.query);
+
+  response.send(expenses.map(expense => normalizeExpense(expense)));
+};
+
+const getOneExpense = async(request, response) => {
+  const { expenseId } = request.params;
+
+  if (!expenseId) {
+    response.sendStatus(400);
```
33Orange, line 17: Same about the error message in expenses

33Orange (changes requested):
Good job

A few moments with errors

PR #105, opened 2023-08-20
Develop

adv555:


src/models/Category.js
```diff
@@ -0,0 +1,39 @@
...
+        },
+        isUnique(value, next) {
+          Category.findOne({ where: { name: value } }).then((category) => {
+            if (category) {
+              return next('Category already exists');
+            }
+            next();
+          });
+        },
+      },
```
anstsot, line 31: great job with validation

src/models/Category.js
```diff
@@ -0,0 +1,39 @@
+import { DataTypes } from 'sequelize';
+import { sequelize } from '../utils/db.js';
+
+const Category = sequelize.define(
+  'category',
+  {
```
anstsot, line 6: you should have `id` as primary key and use `guid`

src/models/Expense.js
```diff
@@ -0,0 +1,119 @@
+import { sequelize } from '../utils/db.js';
+import { DataTypes } from 'sequelize';
+import { User } from './User.js';
+import { Category } from './Category.js';
+
+const Expense = sequelize.define(
+  'expense',
+  {
```
anstsot, line 8: you should have `id` as primary key and use `guid`

src/models/User.js
```diff
@@ -0,0 +1,30 @@
+import { sequelize } from '../utils/db.js';
+import { DataTypes } from 'sequelize';
+
+const User = sequelize.define(
+  'user',
+  {
```
anstsot, line 6: you should have `id` as primary key and use `guid`

src/services/expenses.service.js
```diff
@@ -0,0 +1,74 @@
...
+
+  if (to) {
+    where.spentAt = { [Op.lte]: to };
+  }
+
+  return Expense.findAll({
+    where,
+    attributes: { exclude: ['createdAt', 'updatedAt', 'deletedAt'] },
+    order: [['spentAt', 'DESC']],
+  });
```
anstsot, line 29: great job with creating complicated query

src/setup.js
anstsot, line 1: great job! if you want to create some data on DB, just use `seeds` ([here](https://sequelize.org/docs/v6/other-topics/migrations/) is info)

anstsot (approved):
Great work, approving:sparkles:
Pay attention to my comments, please

src/models/Category.js
adv555: but I have it already 😊 
I was using the default id primary key property given by Sequelize ( integer, autoincrement). Same for other models
Screenshot 2023-08-23 at 13 58 53

PR #106, opened 2023-08-21
solution

MariiaMonakhova:


src/services/categories.js
```diff
@@ -0,0 +1,67 @@
+'use strict';
+
+const { DataTypes } = require('sequelize');
+const { sequelize } = require('../utils/db.js');
+
+const Category = sequelize.define('Category',
+  {
+    id: {
+      type: DataTypes.INTEGER,
```
anstsot, line 9: better to use `guid` for primary keys

src/createServer.js
```diff
@@ -0,0 +1,28 @@
...
+User.sync({ force: true });
+Expense.sync({ force: true });
+Category.sync({ force: true });
+
+app.use(cors());
+app.use(express.json());
+
+app.use('/expenses', expensesRouter);
+app.use('/users', usersRouter);
+app.use('/categories', categoriesRouter);
```
anstsot, line 24: better to create `createServer` function and use it further

src/controllers/categories.js
```diff
@@ -0,0 +1,82 @@
...
+  res.send(categories);
+};
+
+const getOne = async(req, res) => {
+  const { categoryId } = req.params;
+  const foundCategory = await categoryService.getById(categoryId);
+
+  if (!foundCategory) {
+    res.statusCode = 404;
+    res.send('Category not found');
```
anstsot, line 17:
```suggestion
    res.status(404).send('Category not found');
```

src/controllers/expenses.js
```diff
@@ -0,0 +1,114 @@
...
+  const foundExpense = await expenseService.getById(expenseId);
+
+  if (!foundExpense) {
+    res.statusCode = 404;
+    res.send('Expense is not found');
+
+    return;
+  }
+
+  const { body } = req;
```
anstsot, line 84: better to check the body to not propagate wrong data and it's necessary to check if any info in `body` exists and throw error if there are no data

src/controllers/users.js
```diff
@@ -0,0 +1,82 @@
...
+  const foundUser = userService.getById(userId);
+
+  if (!foundUser) {
+    res.statusCode = 404;
+    res.end('User not found');
+
+    return;
+  }
+
+  const { name } = req.body;
```
anstsot, line 53: check if `name` exists and throw error in bad case

anstsot (changes requested):
Good start:four_leaf_clover:
Let’s add a few fixes to improve your code.

src/models/expenses.js
```diff
@@ -0,0 +1,45 @@
...
+  {
+    id: {
+      type: DataTypes.UUID,
+      defaultValue: UUIDV4,
+      primaryKey: true,
+    },
+    userId: {
+      type: DataTypes.UUID,
+      allowNull: false,
+    },
```
anstsot, line 16: set the relation to `User` model

src/services/expenses.js
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  if (from && to) {
+    expensesFromDB = expensesFromDB.filter((exp) => {
+      return exp.spentAt >= from && exp.spentAt <= to;
+    });
+  }
+
+  return expensesFromDB;
+};
```
anstsot, line 27: better to create complicated query instead

anstsot (approved):
But pay attention to my comments to avoid the same mistakes in the future

PR #107, opened 2023-08-22
add task solution

ollavka:


src/utils/db.js
```diff
@@ -0,0 +1,6 @@
+import { Sequelize } from 'sequelize';
+
+export const sequelize = new Sequelize('postgres', 'postgres', 'qwerty1234', {
```
anstsot, line 3: better to create constants
```suggestion
const DATABASE = 'postgres';
const USERNAME = 'postgres';
const PASSWORD = 'qwerty1234';

export const sequelize = new Sequelize(DATABASE, USERNAME, PASSWORD, {
```

src/services/expenses.js
```diff
@@ -0,0 +1,82 @@
...
+    });
+  }
+
+  return filteredExpenses;
+}
+
+export async function getAll(searchQuery) {
+  const initialExpenses = await Expense.findAll();
+
+  return getExpensesByQuery(initialExpenses, searchQuery);
```
anstsot, line 46: it will be better to try use `sequlize` to filter the data

src/controllers/expenses.js
```diff
@@ -0,0 +1,119 @@
...
+  const expenseData = req.body;
+
+  const expectedFields = [
+    'userId',
+    'spentAt',
+    'title',
+    'amount',
+    'category',
+    'note',
+  ];
```
anstsot, line 37: you can get them from model
```suggestion
  const expectedFields = Expense.getAttributes();
```

anstsot (approved):
But pay attention to my comments to avoid the same mistakes in the future

PR #110, opened 2023-08-24
Develop

AlyonaSarapina:


src/controllers/expenses.js
```diff
@@ -0,0 +1,130 @@
...
+} = require('../services/expenses');
+
+const getAllExpenses = async (req, res) => {
+  const { userId, from, to, categories } = req.query;
+  let preparedExpenses = await getAll();
+
+  if (userId) {
+    preparedExpenses = preparedExpenses
+      .filter(expense => expense.userId === +userId);
+  }
```
artkas03, line 21: In 20 line you can avoid converting `userId` several times for each iteration. You can make separate initialization for `userId` and make it number instantly 
`const userId = +req.query.userId`

artkas03: It's ok, but also you could use sequelize to filter your models but not array methods
If you are interested look through [this](https://sequelize.org/docs/v6/core-concepts/model-querying-basics/#applying-where-clauses)

src/controllers/expenses.js
```diff
@@ -0,0 +1,130 @@
...
+
+  const foundExpense = await getOne(expenseId);
+
+  if (!foundExpense) {
+    res.sendStatus(404);
+
+    return;
+  }
+
+  res.send(normalize(foundExpense));
```
artkas03, line 58: It would be great if you send status here too.
You could chaing this methods `res.sendStatus(statusCode).send(info)`
for example here your statusCode can be 200 and info can be normalized object

src/controllers/users.js
```diff
@@ -0,0 +1,95 @@
...
+  create,
+  remove,
+  update,
+  normalize,
+} = require('../services/users');
+
+const getAllUsers = async (req, res) => {
+  const users = await getAll();
+
+  res.send(users.map(user => normalize(user)));
```
artkas03, line 16: Here you can pass to the map just `normalized ` function because map gets a callback which it executes for each item.
`users.map(normalize);`

artkas03 (approved):
GJ)
Added a few tips for you

src/utils/db.js
```diff
@@ -0,0 +1,12 @@
+'use strict';
+
+const { Sequelize } = require('sequelize');
+
+const sequelize = new Sequelize('postgres', 'postgres', '1234', {
```
anstsot, line 5: better to create constants
```suggestion
const DB_NAME = 'postgres';
//...

const sequelize = new Sequelize(DB_NAME, 'postgres', '1234', {
```

src/models/Expense.js
```diff
@@ -0,0 +1,46 @@
...
+  id: {
+    type: DataTypes.UUID,
+    defaultValue: DataTypes.UUIDV4,
+    primaryKey: true,
+  },
+  userId: {
+    type: DataTypes.UUID,
+    allowNull: false,
+    field: 'user_id',
+  },
```
anstsot, line 16: set the relation to User model

src/controllers/expenses.js
```diff
@@ -0,0 +1,130 @@
...
+
+    preparedExpenses = preparedExpenses.filter(expense =>
+      new Date(expense.spentAt) < toDate
+    );
+  }
+
+  if (categories) {
+    preparedExpenses = preparedExpenses
+      .filter(expense => expense.category === categories);
+  }
```
anstsot, line 42: сreate complicated query instead

src/controllers/expenses.js
```diff
@@ -0,0 +1,130 @@
...
+    spentAt,
+    title,
+    amount,
+    category,
+    note,
+  });
+
+  res.statusCode = 201;
+
+  res.send(newExpense);
```
anstsot, line 90:
```suggestion
  res.status(201).send(newExpense);
```

src/controllers/expenses.js
```diff
@@ -0,0 +1,130 @@
...
+
+  if (!expense) {
+    res.sendStatus(404);
+
+    return;
+  }
+
+  const { body } = req;
+
+  await update(expenseId, body);
```
anstsot, line 119: better to check if `body` is correct and throw error in case it's not

anstsot (changes requested):
Good start:four_leaf_clover:
Let’s add a few fixes to improve your code.

PR #113, opened 2023-08-26
Develop

ilona-borkovska:


src/utils/filterExpenses.js
anstsot, line 1: it will be great to create complicated query to DB instead of filtering

PR #114, opened 2023-08-27
node_accounting-app-with-db solution

NataliGru:


src/services/expense.service.js
```diff
@@ -0,0 +1,63 @@
...
+
+  if (from && to) {
+    queryParameters.spentAt = {
+      [Op.between]: [from, to],
+    };
+  }
+
+  const expenses = await Expense.findAll({
+    where: queryParameters,
+  });
```
anstsot, line 30: great job with creating a complicated query!

src/utils/db.js
```diff
@@ -0,0 +1,13 @@
+'use strict';
+
+const { Sequelize } = require('sequelize');
+
+const sequelize = new Sequelize('postgres', 'postgres', 'natali2023', {
```
anstsot, line 5: better to create constants
```suggestion
const DB_NAME = 'postgres';
// ...

const sequelize = new Sequelize(DB_NAME, 'postgres', 'natali2023', {
```

anstsot (approved):

PR #116, opened 2023-08-29
Solution

Ilia991:


src/models/users.js
```diff
@@ -0,0 +1,14 @@
+const { DataTypes } = require("sequelize");
+const sequelize = require("../utils/db.js");
+
+const Users = sequelize.define('Users', {
```
maxim2310, line 4: you need to set up id for user like or use uiid type
image

src/models/expenses.js
```diff
@@ -0,0 +1,33 @@
+const { DataTypes } = require("sequelize");
+const sequelize = require("../utils/db.js");
+
+const Expenses = sequelize.define('Expenses', {
```
maxim2310, line 4: also here you must to set up id and make assign dependency on user table 
image

src/api/users.json
```diff
@@ -0,0 +1,10 @@
+[
+  {
+    "id": 1,
+    "name": "Bob"
+  },
+  {
+    "id": 2,
+    "name": "Bib"
+  }
+]
```
maxim2310, line 10: why do you need it?

src/controllers/expenses.js
```diff
@@ -0,0 +1,100 @@
...
+};
+
+const createExpense = (req, res) => {
+  const { userId, amount, category, note } = req.body;
+
+  if (!userId || !amount || !category || !note) {
+    res.sendStatus(400);
+  }
+
+  const newExpences = create({
```
maxim2310, line 44: you need to await creating of user

src/controllers/expenses.js
```diff
@@ -0,0 +1,100 @@
...
+    userId, amount, category, note,
+  });
+
+  res.sendStatus(201);
+  res.end(newExpences);
+};
+
+const getExpense = (req, res) => {
+  const { id } = req.params;
+  const foundExpenses = getById(id);
```
maxim2310, line 54: also here need to await

src/services/expensesServices.js
```diff
@@ -0,0 +1,50 @@
+'use strict';
+
+const { Sequelize } = require("sequelize");
+const Expenses = require("../models/expenses");
+const sequelize = require("../utils/db");
```
maxim2310, line 5: you dont use it here

src/services/expensesServices.js
```diff
@@ -0,0 +1,50 @@
+'use strict';
+
+const { Sequelize } = require("sequelize");
+const Expenses = require("../models/expenses");
+const sequelize = require("../utils/db");
+
+
+async function getAll() {
+  return await Expenses.findAll();
+};
```
maxim2310, line 10: this func could be not async and not return await

maxim2310 (changes requested):
Good job! Let's improve your code a bit)
check all pleases where you use async func you need to await response and catch probably error

src/utils/db.js
```diff
@@ -0,0 +1,17 @@
+'use strict';
+
+require('dotenv/config');
+
+const { Sequelize } = require('sequelize');
+
+const sequelize = new Sequelize(
+  'postgres',
+  'postgres',
+  '3562351Boh',
```
lerastarynets, line 10: move this to .env file pls, if u don't understand something, ask for help in **fe_chat**

lerastarynets (changes requested):
Great job overall! Just one tiny thing needs to be fixed before approval)

src/utils/db.js
Ilia991: I had everything in the env file, but it didn’t connect to the postrges server through it, but everything works in another task

lerastarynets (approved):
You've got this, keep pushing forward!

PR #118, opened 2023-09-01
Solution

Kinderwagen:


src/services/userServices.js
```diff
@@ -0,0 +1,42 @@
...
+const User = require('../models/user');
+
+const getAll = async() => {
+  const users = await User.findAll();
+
+  return users;
+};
+
+const getUserById = async(userId) => {
+  return User.findOne({ where: { id: userId } });
```
DarkMistyRoom, line 12: you can use `findByPk` method

DarkMistyRoom (approved):
Good job!

PR #119, opened 2023-09-04
Solution

VolodymyrPst:


src/controllers/expenses.js
```diff
@@ -0,0 +1,118 @@
...
+  const allowedKeys = [
+    'title',
+    'amount',
+    'category',
+    'note',
+    'userId',
+  ];
+  const keysExpense = Object.keys(expense);
+
+  const missingFields = keysExpense.every(key => allowedKeys.includes(key));
```
anastasiiavorobiova, line 33: Consider adding middleware for validation

```suggestion

```

src/controllers/user.js
```diff
@@ -0,0 +1,93 @@
...
+} = require('../services/userService');
+
+async function getAllUsersAction(req, res) {
+  try {
+    const users = await getAllUsers();
+
+    res.send(users.map(user => normalize(user)));
+  } catch (error) {
+    res.sendStatus(400).send(error.message);
+  }
```
anastasiiavorobiova, line 19:
```suggestion
  try {
    const users = await getAllUsers();
    
    if(no users) {
      res.sendStatus(400).send(error text);
    }

    res.send(users.map(user => normalize(user)));
  } catch (error) {
    res.sendStatus(500).send(generic error message);
  }
```

src/controllers/user.js
```diff
@@ -0,0 +1,93 @@
...
+};
+
+async function addUserAction(req, res) {
+  const { name } = req.body;
```
anastasiiavorobiova, line 23: Add middleware here

anastasiiavorobiova (changes requested):
Great work! Added comments to fix

anastasiiavorobiova (approved):
Good work

PR #121, opened 2023-09-15
Solution 1

JohnRust12:


.env
```diff
@@ -0,0 +1,4 @@
+NAME=postgres
```
witflash, line 1: It's better to add `.env` file to `.gitignore` list.
Any `env` files should be only your local files.

witflash (approved):
Well done 👍

PR #122, opened 2023-09-15
added solution

sheva10barca:


src/controllers/expenses.js
```diff
@@ -0,0 +1,119 @@
...
+      const normalizedDateFrom = new Date(from);
+      const normalizedDateTo = new Date(to);
+
+      return (
+        normalizedExpenseDate >= normalizedDateFrom
+        && normalizedExpenseDate <= normalizedDateTo
+      );
+    });
+  }
+
```
lerastarynets, line 32: consider moving all the complex logic to service file as controller is just a middleware between service and route so it shouldn't have this logic

lerastarynets (approved):
Great job overall! Left one tiny suggestion for the improvement

PR #140, opened 2023-10-23
only a draft for now

AndrzejPec:
I'm confused, most likely I have misunderstood the readme content or the purpose of the task whatsoever.
Perhaps we need to follow the video guides... --TODO

### DIFF
--- a/src/models/categoryModel.js
+++ b/src/models/categoryModel.js
@@ -1,24 +1,23 @@
-'use strict';
+// IM NOT SURE IF THIS MIGHT BE NEEDED IN THE FUTURE SO I'M KILLING IT FOR NOW
 
-const pool = require('../dbConfig');
+// 'use strict';
 
-const getCategories = async() => {
-  const res = await pool.query('SELECT * FROM categories');
+// const pool = require('../dbConfig');
 
-  return res.rows;
-};
+// const getCategories = async() => {
+//   const res = await pool.query('SELECT * FROM categories');
 
-const createCategory = async(name) => {
-  const res = await pool
-    .query('INSERT INTO categories (name) VALUES ($1) RETURNING *', [name]);
+//   return res.rows;
+// };
 
-  return res.rows[0];
-};
+// const createCategory = async(name) => {
+//   const res = await pool
+//     .query('INSERT INTO categories (name) VALUES ($1) RETURNING *', [name]);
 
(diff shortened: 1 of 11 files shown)


src/models/categoryModel.js
choeqq, line 1: since you've already done basically the same in `dbQueries` file, I think you can drop this one

src/routes/categoryRoutes.js
```diff
@@ -0,0 +1,143 @@
+'use strict';
+
+const express = require('express');
+const {
+  createCategoryService,
+  getCategoryByIdService,
+  updateCategoryService,
+  deleteCategoryService,
+} = require('../services/categoryService');
```
choeqq, line 9: I think it would be ideal to export the whole service with methods on it, just to access them as follows:
`categoryService.createCategory`

src/routes/categoryRoutes.js
```diff
@@ -0,0 +1,143 @@
...
+//       return res.status(404).send();
+//     }
+
+//     res.status(204).send();
+//   });
+
+//   return router;
+// };
+
+// module.exports = categoryRoutes;
```
choeqq, line 143: I don't think it will be, you always have GIT history, just in case

src/services/categoryService.js
```diff
@@ -0,0 +1,83 @@
...
+  return getCategoryById(id);
+};
+
+const updateCategoryService = async(id, name) => {
+  return updateCategory(id, name);
+};
+
+const deleteCategoryService = async(id) => {
+  return deleteCategory(id);
+};
```
choeqq, line 24: prettier?

src/services/categoryService.js
```diff
@@ -0,0 +1,83 @@
...
+
+//   return true;
+// };
+
+// module.exports = {
+//   createCategory,
+//   getCategoryById,
+//   updateCategory,
+//   deleteCategory,
+// };
```
choeqq, line 83: I don't think it will be, you always have GIT history, just in case

choeqq (changes requested):

src/services/categoryService.js
AndrzejPec: it may look identical because adding spaces between async and (param) is all the formatter did here

choeqq (approved):

PR #141, opened 2023-10-25
Solution

HaniaNassalska:


src/expenses/expenses.service.js
```diff
@@ -0,0 +1,129 @@
...
+  },
+  category: {
+    type: DataTypes.STRING,
+  },
+  note: {
+    type: DataTypes.STRING,
+  },
+}, {
+
+});
```
choeqq, line 36: that could be extracted into `/expenses.model.js` file, just so you know 😄

src/users/users.service.js
```diff
@@ -0,0 +1,90 @@
...
+  },
+  id: {
+    type: DataTypes.UUID,
+    defaultValue: DataTypes.UUIDV4,
+    primaryKey: true,
+  },
+}, {
+});
+
+Users.sync();
```
choeqq, line 21: that could also be extracted into `/users.model.js` just so you know 😄

main.js
```diff
@@ -0,0 +1,9 @@
+'use strict';
+
+const { createServer } = require('./src/createServer');
+
+createServer()
+  .listen(3000, () => {
```
choeqq, line 6: since you're already using `.env` you could pass port to `.listen` method, it is popular practice 😄

choeqq (approved):
Great solution - approved, please just take a look at my comments 😄 

@HaniaNassalska

PR #142, opened 2023-10-26
Solution

jakubMikolajczykTech:


src/models/Expense.js
choeqq, line 1: I think you've accidently put router inside models, please make sure to replace it with Expenses model

src/utils/db.js
```diff
@@ -0,0 +1,12 @@
+'use strict';
+
+const { Sequelize } = require('sequelize');
+
+const sequelize = new Sequelize('postgres', 'postgres', 'password', {
```
choeqq, line 5: It would be great if those came from .env file, but it's not mandatory

src/server.js
```diff
@@ -0,0 +1,12 @@
...
+const express = require('express');
+const { userRouter } = require('./routes/users');
+const { expenseRouter } = require('./routes/expenses');
+
+const server = express();
+
+server.use('/users', userRouter);
+server.use('/expenses', expenseRouter);
+
+server.listen(3000);
```
choeqq, line 12: if you'll decide to create .env file you could also add env variable for PORT and pass it here

choeqq (changes requested):

PR #143, opened 2023-10-26
Develop

anpiet:


main.js
```diff
@@ -0,0 +1,14 @@
...
+
+const dotenv = require('dotenv');
+
+dotenv.config();
+
+const { createServer } = require('./src/createServer.js');
+
+const PORT = process.env.PORT || 8080;
+
+createServer().listen(3000, () => {
```
choeqq, line 11: You've already put so much effor in getting PORT from env, please use it here instead of 3000 😄

src/services/database.service.js
```diff
@@ -0,0 +1,14 @@
+'use strict';
+
+const Sequelize = require('sequelize');
+const dotenv = require('dotenv');
+
+dotenv.config();
+
+const sequelize = new Sequelize(process.env.EXT_URL, {
```
choeqq, line 8: could you please define some default URL, when env EXT URL isn't passed?

choeqq (changes requested):
Looks great, just minor improvements

choeqq (approved):
LGTM ✅

PR #146, opened 2023-11-21
solution

UlianaChorna:


src/controllers/expense.controller.js
```diff
@@ -0,0 +1,132 @@
+'use strict';
+
+const expenseService = require('../services/expenses.service');
+const { getUserById } = require('../services/user.service');
+
+const getExpenses = async(req, res) => {
```
lerastarynets, line 6: controllers are just a middleware between route and service and they r supposed to have as least logic as possible, so pls consider moveing the majority of logic to services

src/controllers/user.controller.js
```diff
@@ -0,0 +1,87 @@
...
+  }
+
+  if (typeof name !== 'string') {
+    res.sendStatus(422);
+
+    return;
+  }
+
+  await userService.updateUser({
+    id: id, name,
```
lerastarynets, line 73:
```suggestion
    id, name,
```

src/services/expenses.service.js
```diff
@@ -0,0 +1,64 @@
...
+    userId, spentAt, title, amount, category, note,
+  });
+};
+
+const update = ({
+  id,
+  data,
+}) => {
+  return Expense.update(data, {
+    where: { id: id },
```
lerastarynets, line 45:
```suggestion
    where: { id },
```
**PLS DO THE SAME FOR ALL SUCH CASES IN THE PROJECT**

lerastarynets (approved):
Great job overall! Left a few comments to make your code more advanced)

PR #147, opened 2023-11-23
solution

vlkzmn:


src/controllers/expensesController.js
```diff
@@ -0,0 +1,188 @@
...
+      expenses = await expensesService.getAll();
+
+      res.send(expenses);
+    } catch (error) {
+      res.sendStatus(500);
+    }
+  } else {
+    res.sendStatus(400);
+  }
+};
```
Andriy-Shymkiv, line 73: to much logic inside controller, it's not a correct approach. you should move this to service, because the main task of controller is to take request and send response. All business logic should be inside service

src/controllers/expensesController.js
```diff
@@ -0,0 +1,188 @@
...
+
+const create = async(req, res) => {
```
Andriy-Shymkiv, line 75: the same about each functions in this controller

src/createServer.js
```diff
@@ -0,0 +1,20 @@
...
+const { expensesRouter } = require('./routes/expensesRoute');
+const usersService = require('./services/usersService');
+const expensesService = require('./services/expensesService');
+
+const app = express();
+
+usersService.clearUsers();
+expensesService.clearExpenses();
+
+app.use('/users', express.json(), usersRouter);
```
Andriy-Shymkiv, line 15: use `express.json()` middleware once for each endpoint

src/services/createTables.js
```diff
@@ -0,0 +1,7 @@
+'use strict';
+
+const { User } = require('./usersService');
+const { Expense } = require('./expensesService');
+
+User.sync({ force: true });
+Expense.sync({ force: true });
```
Andriy-Shymkiv, line 7: It shouldn't be a separate service for this logic, just put it into the users/expenses services

src/services/db.js
```diff
@@ -0,0 +1,10 @@
+'use strict';
+
+const { Sequelize } = require('sequelize');
+
+const sequelize = new Sequelize('postgres', 'postgres', '123123', {
```
Andriy-Shymkiv, line 5: for now it is ok, but in real projects you should put this variables in env

Andriy-Shymkiv (changes requested):

src/controllers/expensesController.js
vlkzmn: https://en.wikipedia.org/wiki/Model%E2%80%93view%E2%80%93controller
The controller responds to the user input and performs interactions on the data model objects. The controller receives the input, optionally validates it and then passes the input to the model.

src/services/createTables.js
vlkzmn: Why is this in the services? This file is run once when creating tables, and then it is better to delete it to avoid resetting tables. It is also called table creation.

src/services/db.js
vlkzmn: That's what we'll do in the next thread ))

src/controllers/expensesController.js
lerastarynets: @vlkzmn the previous mentor is right, controllers should have as little logic as possible, they are just middleware between route and service, so pls consider moving the majority of logic to the services
[see the article](https://www.coreycleary.me/what-is-the-difference-between-controllers-and-services-in-node-rest-apis)

lerastarynets (approved):
Great job overall! Left one tiny suggestion for the improvement

PR #148, opened 2023-11-25
Solution

OlegPopovych:
Hello!


src/controllers/users.controller.js
```diff
@@ -0,0 +1,119 @@
...
+
+    res.send(result.map(user => usersService.normalizeUserData(user)));
+  } catch (err) {
+    console.error('Error executing query:', err);
+
+    res.status(500).send(err.message);
+  }
+};
+
+const getParticular = async(req, res) => {
```
vasylmiskiv, line 19: getUser or getUserById would be much appropriate name of the function
```suggestion
const getUser = async(req, res) => {
```

src/controllers/expenses.controller.js
```diff
@@ -0,0 +1,146 @@
+'use strict';
+
+/* eslint no-console: ["error", { allow: ["warn", "error"] }] */
+
+const expensesService = require('../services/expenses.service');
+const usersService = require('../services/user.service');
+const { Expense } = require('../controllers/db/models/expense.model');
+
+const getById = async(req, res) => {
```
vasylmiskiv, line 9:
```suggestion
const getExpense= async(req, res) => {
```

vasylmiskiv (approved):
Looks fine but pay attention to the comments

PR #157, opened 2023-12-10
accounting app with db

ElenaGrasovskaya:
First, run `node src/setup.js` to create BD tables,
Then run `node main.js` to start the server.

fingers crossed...


mpohorenyi (approved):
LGTM 🔥

PR #158, opened 2023-12-11
Solution_v1.0

Rozdorozhnii:


src/index.js
```diff
@@ -1 +1,22 @@
 'use strict';
+
+const express = require('express');
+
+require('dotenv').config();
+
+const { usersRouter } = require('./routes/users.route');
+const { expensesRouter } = require('./routes/expenses.route');
+
+const PORT = process.env.API_PORT;
```
Andriy-Shymkiv, line 10: what if `process.env.API_PORT` is undefined?
```suggestion
const PORT = process.env.API_PORT ?? 3001;
```

src/services/expenses.services.js
```diff
@@ -0,0 +1,106 @@
...
+}, {
+  tableName: 'expenses',
+});
+
+const normalizeExpense
+  = ({ id, userId, spentAt, title, amount, category, note }) => {
+    return {
+      id, userId, spentAt, title, amount, category, note,
+    };
+  };
```
Andriy-Shymkiv, line 44: what's the purpose of this function? there is no business logic inside

Andriy-Shymkiv (changes requested):
good job! just a few moments to improve

src/services/expenses.services.js
Rozdorozhnii: this function need to excluding unnecessary data from database

src/index.js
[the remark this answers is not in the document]
Rozdorozhnii: image
`index.js` isn't in `controllers` folder, probably you made a mistake

src/index.js
Rozdorozhnii: Unfortunately I can use logical OR `||` only due linter setup

src/services/expenses.services.js
```diff
@@ -0,0 +1,106 @@
+'use strict';
+
+const { DataTypes } = require('sequelize');
+const { Op } = require('sequelize');
+
+const { sequelize } = require('./database.service');
+
+const Expense = sequelize.define('Expense', {
```
denys-danyliuk, line 8: would be better to move model to other file

denys-danyliuk (approved):
Good job!

PR #161, opened 2023-12-20
accounting app with database in postgres

sofiiaTriasko:


src/controllers/expenses.controllers.js
```diff
@@ -0,0 +1,103 @@
...
+  const expenses = expensesService.getExpenses({
+    userId, categories, from, to,
+  });
+
+  if (!expenses) {
+    res.sendStatus(404);
+
+    return;
+  }
+  res.send((await expenses).map(expense => expensesService.normalize(expense)));
```
denys-danyliuk, line 17: probably would be better to move this to service cause you are kinda interacting with a data itself, controllers just receive data from the user sends it to service, receives data from him and sends it to user

src/controllers/expenses.controllers.js
```diff
@@ -0,0 +1,103 @@
...
+  const expense = expensesService.getExpenseById(id);
+
+  if (!expense) {
+    res.sendStatus(404);
+
+    return;
+  }
+
+  expensesService.deleteExpense(expense).then(() => {
+    // eslint-disable-next-line no-console
```
denys-danyliuk, line 44: you can disable console for whole file to avoid using this rule couple more times in the code

src/controllers/expenses.controllers.js
```diff
@@ -0,0 +1,103 @@
...
+    console.log('Expense was added successfully');
+  })
+    .catch((error) => {
+      // eslint-disable-next-line no-console
+      console.error('Error occurred:', error);
+    });
+
+  res.statusCode = 201;
+  res.send(204);
```
denys-danyliuk, line 72: why do you send statusCode and then 204? you should send back created entity

src/services/expenses.service.js
```diff
@@ -0,0 +1,123 @@
+'use strict';
+
+const { User } = require('./users.service.js');
+const { DataTypes, Op } = require('sequelize');
+const { sequelize } = require('../db');
+const { uuidv4 } = require('uuid');
+
+const Expense = sequelize.define('Expense', {
```
denys-danyliuk, line 8: you can move models to separate file but it's up to you

src/services/expenses.service.js
```diff
@@ -0,0 +1,123 @@
...
+          [Op.lte]: from,
+        },
+      },
+    });
+  }
+};
+
+const getExpenseById = (id) => {
+  return Expense.findByPk(id);
+};
```
denys-danyliuk, line 90:
```suggestion
const getExpenseById = (id) => {
  return Expense.findByPk(id);
};

```

src/services/users.service.js
```diff
@@ -0,0 +1,64 @@
...
+const { sequelize } = require('../db');
+const { uuidv4 } = require('uuid');
+const { DataTypes } = require('sequelize');
+
+const User = sequelize.define('User', {
+  name: {
+    type: DataTypes.STRING,
+    allowNull: false,
+  },
+},
```
denys-danyliuk, line 12: it also should have id, doesn't it?

denys-danyliuk (changes requested):
Good job!
Pay attention to the comments

src/services/users.service.js
sofiatryasko: my app make table from ORM model using sync force so according to video lessons id will be made automatically, is it correct?

denys-danyliuk: yes, it is, I just thought that you've created table at first and then created model

denys-danyliuk (approved):
Good job!🔥

PR #165, opened 2023-12-29
Develop

Sergey-Mironenko:


src/services/expensesService.js
```diff
@@ -0,0 +1,71 @@
...
+  = (expenseId, userId, spentAt, title, amount, category, note = 'empty') => {
+    client.query(`
+    UPDATE expenses
+    SET userId = $1,
+      spentAt = $2,
+      title = $3,
+      amount = $4,
+      category = $5,
+      note = $6
+    WHERE id = $7
```
loralevitska, line 42: Use incoming variables, fix all cases if present

loralevitska (approved):
Almost done) Let's improve

PR #167, opened 2024-01-07
 Accounting_app_using_PostgreSQL

rialleons:
Also added category check on creating and updating expenses (if category does not exist the new one is added to the "categories" table).

### DIFF
--- a/src/services/user.service.js
+++ b/src/services/user.service.js
@@ -1,0 +1,67 @@
+'use strict';
+
+const { DataTypes } = require('sequelize');
+
+const { sequelize } = require('./db.js');
+
+const User = sequelize.define('User', {
+  id: {
+    type: DataTypes.INTEGER,
+    autoIncrement: true,
+    primaryKey: true,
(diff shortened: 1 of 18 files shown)


src/services/user.service.js
```diff
@@ -0,0 +1,67 @@
...
+  },
+  name: {
+    type: DataTypes.STRING,
+    allowNull: false,
+  },
+}, {
+  tableName: 'users',
+  updatedAt: false,
+  createdAt: false,
+});
```
DarkMistyRoom, line 21: move it to a separate file. The same for all models

src/services/user.service.js
```diff
@@ -0,0 +1,67 @@
...
+
+const getAll = async() => {
+  const result = await User.findAll({
+    order: ['id'],
```
DarkMistyRoom, line 25: better use createdAt for the order

DarkMistyRoom (approved):
Good job!

PR #168, opened 2024-01-07
result

errisi:


src/services/expenses.service.js
```diff
@@ -0,0 +1,107 @@
+'use strict';
+
+const { DataTypes } = require('sequelize');
+const sequelize = require('./../database');
+
+const Expense = sequelize.define('Expense', {
+  id: {
```
BudnikOleksii, line 7: You shouldn't define schemas in services. Create a separate file for you model

src/services/expenses.service.js
```diff
@@ -0,0 +1,107 @@
...
+
+  if (categories) {
+    filteredExpenses = filteredExpenses.filter((expense) =>
+      categories.includes(expense.category)
+    );
+  }
+
+  if (from && to) {
+    filteredExpenses = filteredExpenses.filter(
+      (expense) =>
```
BudnikOleksii, line 58: Why do you query all expenses and then filter them instead of querying only what you actually need from the database?

src/services/expenses.service.js
```diff
@@ -0,0 +1,107 @@
...
+    attributes: ['id'],
+    order: [['id', 'DESC']],
+  });
+
+  const getNewExpenseId = lastExpense ? lastExpense.id + 1 : 0;
+
+  const newExpense = {
+    ...expense,
+    id: getNewExpenseId,
+  };
```
BudnikOleksii, line 80: Do you need this? The id should be generated automatically

src/services/expenses.service.js
```diff
@@ -0,0 +1,107 @@
...
+
+  Expense.create(newExpense);
```
BudnikOleksii, line 82: I'm not sure, do you return a new entity?

src/services/users.service.js
```diff
@@ -0,0 +1,67 @@
+'use strict';
+
+const { DataTypes } = require('sequelize');
+const sequelize = require('./../database');
+
+const User = sequelize.define('User', {
```
BudnikOleksii, line 6: The same thing, services only for the business logic

src/services/users.service.js
```diff
@@ -0,0 +1,67 @@
...
+    attributes: ['id'],
+    order: [['id', 'DESC']],
+  });
+
+  const getNewUserId = lastUser ? lastUser.id + 1 : 0;
+
+  const newUser = {
+    id: getNewUserId,
+    name,
+  };
```
BudnikOleksii, line 40: The same

BudnikOleksii (changes requested):
Good work 👍 
Left a few comments, check them

ArtemTeslenko (approved):
Good job 👍

PR #169, opened 2024-01-09
acc app with DB

Mykola-Hadupiak:
I made a short review
LOOK

https://www.loom.com/share/3ffd93466cd34d3ea5692331310faccb?sid=6944586a-6d5d-4cbf-8c14-ee3790ef7d53

### DIFF
--- a/src/controllers/expenses.controller.js
+++ b/src/controllers/expenses.controller.js
@@ -1,0 +1,129 @@
+/* eslint-disable no-console */
+'use strict';
+
+const { expensesService } = require('../services/expenses.service.js');
+const { userService } = require('../services/user.service.js');
+
+const get = async(req, res) => {
+  const expenses = await expensesService.getAll(req.query);
+
+  res.send(expenses);
+};
+
+const getOne = async(req, res) => {
+  const { id } = req.params;
+
+  const expense = await expensesService.getById(id);
+
+  if (!expense) {
+    res.sendStatus(404);
+
+    return;
+  }
+
+  res.send(expense);
+};
+
+const create = async(req, res) => {
+  const {
+    userId,
+    spentAt,
+    title,
+    amount,
+    category,
+    note,
+  } = req.body;
+
+  if (!userId || !spentAt || !title || amount < 0 || !category || !note) {
+    res.sendStatus(400);
+
+    return;
+  }
+
+  try {
+    await userService.getById(userId);
+  } catch (error) {
+    console.log('User not found');
+    res.sendStatus(400);
+
+    return;
+  }
+
+  if (!(new Date(spentAt) instanceof Date)
+    || typeof title !== 'string'
+    || typeof amount !== 'number'
+    || typeof category !== 'string'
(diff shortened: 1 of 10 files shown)


src/services/expenses.service.js
```diff
@@ -0,0 +1,130 @@
...
+  },
+  note: {
+    type: DataTypes.STRING,
+    allowNull: false,
+  },
+}, {
+  tableName: 'expenses',
+  createdAt: false,
+  updatedAt: false,
+});
```
DarkMistyRoom, line 45: better move the model to a separate file

DarkMistyRoom: the same for the user

src/services/expenses.service.js
```diff
@@ -0,0 +1,130 @@
...
+    if (to) {
+      const toDate = new Date(to);
+
+      if (new Date(expense.spentAt) > toDate) {
+        return false;
+      }
+    }
+
+    return true;
+  });
```
DarkMistyRoom, line 89: better use `where` option to fetch only needed data

DarkMistyRoom (changes requested):
Good job! Only a few suggestions

src/controllers/expenses.controller.js
```diff
@@ -0,0 +1,129 @@
...
+  const expense = await expensesService.getById(id);
+
+  if (!expense) {
+    res.sendStatus(404);
+
+    return;
+  }
+
+  res.send(expense);
+};
```
Andriy-Shymkiv, line 25: add try/catch blocks

src/controllers/expenses.controller.js
```diff
@@ -0,0 +1,129 @@
...
+
+    return;
+  }
+
+  const expense = await expensesService.create(req.body);
+
+  res.statusCode = 201;
+
+  res.send(expense);
+};
```
Andriy-Shymkiv, line 68: to much logic in controller. it should be responsible only for taking data from request and put it in service (all validation, business logic, etc. should be handled  by service), like this:
```js
const create (req,res) => {
  try {
     const expense = await expensesService.create(req.body);
     res.statusCode = 201;
  } catch (e) {
    console.error(`error creating expense: ${e.message}`)
  }
}
```

Andriy-Shymkiv: please fix it everywhere

src/services/db.js
```diff
@@ -0,0 +1,15 @@
+'use strict';
+
+const { Sequelize } = require('sequelize');
+
+const USERNAME = 'postgres';
+const PASSWORD = 'postgres';
```
Andriy-Shymkiv, line 6: in real applications you should keep this variables in `env` file

Andriy-Shymkiv (changes requested):

Andriy-Shymkiv (commented):
I also watched the video review about your app that you recorded. good job!

DarkMistyRoom (approved):
Well done!

PR #170, opened 2024-01-11
node_accounting-app-with-db

Viktorianeimesh:


src/db.js
```diff
@@ -0,0 +1,10 @@
+'use strict';
+
+const Sequelize = require('sequelize');
+
+const sequelize = new Sequelize('postgres', 'postgres', '23081988', {
+  host: 'localhost',
+  dialect: 'postgres',
+});
```
Andriy-Shymkiv, line 8: in a real app you you will need to store this credentials in env file

src/services/expenses.service.js
```diff
@@ -0,0 +1,104 @@
...
+    amount,
+    category,
+    note,
+  };
+
+  return Expenses.create(item);
+};
+
+const removeExpenses = async(id) => {
+  await Expenses.destroy({
```
Andriy-Shymkiv, line 87: you can add try/catch block and return `true` on success

src/services/expenses.service.js
```diff
@@ -0,0 +1,104 @@
...
+    where: {
+      id,
+    },
+  });
+};
+
+const updateExpenses = async(id, title) => {
+  await Expenses.update(title, { where: { id } });
```
Andriy-Shymkiv, line 95: if `update` method returning an updated entity, then you can just:
```js
return Expenses.update(title, { where: { id } });
```
and in controller:
```js
const expense = expensesService.update({})
res.send(expense)
```

Andriy-Shymkiv: please check this and fix it everywhere if possible ⬆️

Andriy-Shymkiv (approved):
very good job! left some comments to improve your codebase

PR #175, opened 2024-01-18
solution

BoblykNazar:


db.js
```diff
@@ -0,0 +1,71 @@
+'use strict';
+
+const { Sequelize, DataTypes } = require('sequelize');
+
+const sequelize = new Sequelize('postgres', 'postgres', '4249', {
```
mykhailoyurkovskyi, line 5: You can move password to .env

src/controllers/expense.controller.js
```diff
@@ -0,0 +1,86 @@
...
+    return;
+  }
+
+  res.send(expense);
+};
+
+const createExpense = async(req, res) => {
+  const expense = req.body;
+
+  const user = await userService.getById(expense.userId);
```
mykhailoyurkovskyi, line 36: DB will do it anyway

mykhailoyurkovskyi (approved):
Looks good to me✔

src/services/expense.service.js
```diff
@@ -0,0 +1,67 @@
...
+const getAll = async(userId, categories, from, to) => {
+  const expenses = await Expense.findAll();
+
+  return filterExpense(expenses,
+    {
+      userId,
+      categories,
+      from,
+      to,
+    });
```
BudnikOleksii, line 15: Query from the database only expenses that you actually need instead of querying all and then filtering them.

BudnikOleksii (changes requested):
Good work 👍 
But, query from the database only expenses that you actually need instead of querying all and then filtering them.

BudnikOleksii (approved):

PR #176, opened 2024-01-18
added task solution

Yevhen-Kolosov:
The server is listening on https://accounting-api-l8k3.onrender.com/ (both DB and API should be up).

Request templates from the previous task: [Swaggerhub](https://app.swaggerhub.com/apis/YURIIHOLIUK_1/Accounting-App/1.0.0#/).


BoblykNazar (approved):
Good job!🚀🚀🚀

src/services/expenses.service.js
```diff
@@ -0,0 +1,162 @@
...
+  if (from && to) {
+    filters.spentAt = { [Op.between]: [fromDate, toDate] }
+  }
+
+
+  Object.keys(filters).forEach(key => {
+    if (filters[key] === undefined) {
+      delete filters[key];
+    }}
+  );
```
BudnikOleksii, line 43: You add filters only if they exists, so it looks redundant

BudnikOleksii (approved):
Good Job 🚀

PR #177, opened 2024-01-18
Develop

mykhailoyurkovskyi:


src/repository/db.js
```diff
@@ -0,0 +1,77 @@
...
+const { Sequelize, DataTypes } = require('sequelize');
+
+const sequelize = new Sequelize('postgres', 'postgres', DB_PASSWORD, {
+  host: 'localhost',
+  dialect: 'postgres',
+});
+
+const Expense = sequelize.define('Expense', {
+  id: {
+    type: DataTypes.INTEGER,
```
Anton-Kuchmasov, line 16: seems better to use UUID

src/services/expenses.services.js
```diff
@@ -0,0 +1,114 @@
...
+
+  const whereClause = {};
+
+  if (userId) {
+    whereClause.userId = userId;
+  }
+
+  if (categories) {
+    if (Array.isArray(categories) && categories.length > 0) {
+      whereClause.category = {
```
Anton-Kuchmasov: nice!

src/services/expenses.services.js
```diff
@@ -0,0 +1,114 @@
...
+      await Expense.create(newExpense);
+    }
+  }));
+
+  const allUsers = await Expense.findAll();
+
+  return allUsers;
+};
+
+const clearExpenses = async() => {
```
Anton-Kuchmasov, line 91: it's such a bad practice to have _any_ method which can **destroy** all info from your DB
especially in real projects, bad using this method can cause even a last error in your job =)

Anton-Kuchmasov (approved):
nice job you did!

src/services/expenses.services.js
```diff
@@ -0,0 +1,106 @@
...
+
+    const newExpense = {
+      id: getMaxId + 1,
+      ...expense,
+    };
+
+    return Expense.create(newExpense);
+  } catch (error) {
+    throw error;
+  }
```
BudnikOleksii, line 55: - If you `catch` an error in the controller you don't need to wrap it here
- You can add autoincrement option for the `id` field, so you don't need to do it manually

src/services/users.services.js
```diff
@@ -0,0 +1,80 @@
...
+
+    const newUser = {
+      id: getMaxId + 1,
+      name,
+    };
+
+    return User.create(newUser);
+  } catch (error) {
+    throw error;
+  }
```
BudnikOleksii, line 30: The same. try/catch is redundant, and `id` should be autoincremented

BudnikOleksii (changes requested):
Looks good 👍 
Just a few moments that should be fixed

PR #178, opened 2024-01-18
done task Accounting app with DB

Anton-Kuchmasov:


yefimov77 (approved):
Good Job 👍

package.json
```diff
@@ -1,28 +1,33 @@
...
-    "start": "mate-scripts start",
-    "lint": "mate-scripts lint",
-    "test:only": "mate-scripts test",
-    "update": "mate-scripts update",
-    "postinstall": "npm run update",
-    "test": "npm run lint && npm run test:only"
+    "start": "node main.js",
+    "dev": "nodemon main.js",
+    "test:only": "mocha",
+    "test": "npm run lint && npm run test:only",
```
BudnikOleksii, line 10: If the command is `test` it should test only 👀

src/controllers/expenses.controller.js
```diff
@@ -0,0 +1,138 @@
...
+  const { userId, categories, from, to } = req.query;
+  let searchedExpenses;
+
+  try {
+    searchedExpenses = await getAllExpenses(
+      userId,
+      categories,
+      from,
+      to,
+    );
```
BudnikOleksii, line 22:
```suggestion
  try {
    const searchedExpenses = await getAllExpenses(req.query);
```
- You don't need to use `let` in your case as you don't change this variable
- And you don't need to unpack data from a query as you pass the same into `getAllExpenses`

src/controllers/users.controller.js
```diff
@@ -0,0 +1,90 @@
...
+
+  res.statusCode = 201;
+  res.send(newUser);
+}
+
+async function updateOneUser(req, res) {
+  const { id } = req.params;
+  const { name } = req.body;
+
+  if (typeof name !== 'string' || name.length === 0) {
```
BudnikOleksii, line 56:
```suggestion
  if (typeof name !== 'string' || !name.length) {
```

src/helpers/filterExpenses.js
```diff
@@ -0,0 +1,37 @@
...
+  }
+
+  if (userId) {
+    filteredExpenses = filteredExpenses.filter(
+      expense => expense.userId === +userId
+    );
+  }
+
+  if (from && to) {
+    const toDate = new Date(to);
```
BudnikOleksii, line 19: You don't need to filter expenses, just query from the DB what you actually need

src/services/expenses.service.js
```diff
@@ -0,0 +1,97 @@
...
+  try {
+    const filteredExpenses = filterExpenses(
+      expensesFromDB, categories, userId, from, to
+    );
+
+    return filteredExpenses;
+  } catch (error) {
+    return error;
+  }
+}
```
BudnikOleksii, line 22: - Query from the database only expenses that you actually need instead of querying all and then filtering them.

- `try/catch` in this case is redundant, and it's a bad idea to return an error, as you try to catch it in the `getExpenses` function, but it's impossible if you just return it

src/services/expenses.service.js
```diff
@@ -0,0 +1,97 @@
...
+
+async function getExpenseById(id) {
+  try {
+    const expense = await Expense.findByPk(id);
+
+    return expense;
+  } catch (error) {
+    return error;
+  }
```
BudnikOleksii, line 31: The same, try/catch isn't a good idea in your `services` you can just catch errors in the controllers and just send error info to the client

src/services/expenses.service.js
```diff
@@ -0,0 +1,97 @@
...
+async function removeExpense(id) {
+  try {
+    await Expense.destroy({
+      where: {
+        id,
+      },
+    });
+  } catch (error) {
+    return error;
+  }
```
BudnikOleksii, line 88: Fix it in all places

BudnikOleksii (changes requested):
Good job 👍 
But could be even better, left some suggestions how to do it

roman-mirzoian (approved):
Well done 👍

PR #179, opened 2024-01-18
finish task

yefimov77:


src/services/usersService.js
```diff
@@ -0,0 +1,63 @@
...
+  },
+  name: {
+    type: DataTypes.STRING,
+    allowNull: false,
+  },
+}, {
+  tableName: 'users',
+  updatedAt: false,
+  createdAt: false,
+});
```
emilionila, line 21: i think you shoul move to a different file

src/services/expensesService.js
```diff
@@ -0,0 +1,142 @@
...
+  },
+  note: {
+    type: DataTypes.TEXT,
+    allowNull: false,
+  },
+}, {
+  tableName: 'expenses',
+  updatedAt: false,
+  createdAt: false,
+});
```
emilionila, line 41: maybe move it to a different file as well

src/createServer.js
```diff
@@ -0,0 +1,26 @@
...
+  app.use(express.json());
+  app.use('/users', userRouter);
+  app.use('/expenses', expensesRouter);
+
+  return app;
+};
+
+module.exports = {
+  createServer,
+};
```
emilionila, line 26: cannot find where you run createServer, maybe just forgot to commit the file where  you do createServer.listen(...)

emilionila (approved):
nicely done, check comments 👍

src/services/expensesService.js
BudnikOleksii: > maybe move it to a different file as well

Good point 👍 
Also if you want to use UUID as `id` the sequelize has corresponding type and you can set value by default so ORM will generate ids for you, and you don't need any other libraries https://sequelize.org/docs/v7/models/data-types/#uuids

src/services/expensesService.js
```diff
@@ -0,0 +1,142 @@
...
+
+  if (userId) {
+    allExpenses = allExpenses.filter(
+      expense => expense.userId === +userId
+    );
+  }
+
+  if (from && to) {
+    allExpenses = allExpenses.filter(
+      item => item.spentAt > from && item.spentAt < to);
```
BudnikOleksii, line 60: Query from the database only expenses that you actually need instead of querying all and then filtering them.

src/services/expensesService.js
```diff
@@ -0,0 +1,142 @@
...
+    await Expenses.update(amount, { where: { id } });
+  }
+
+  if (category) {
+    await Expenses.update(category, { where: { id } });
+  }
+
+  if (note) {
+    await Expenses.update(note, { where: { id } });
+  }
```
BudnikOleksii, line 121: Update your entity in the database only once

src/services/usersService.js
BudnikOleksii: > i think you shoul move to a different file

The same comment as above

BudnikOleksii (changes requested):
Good work 👍 
Check the comments and improve your solution

BudnikOleksii (approved):
Nicely done 🚀

PR #181, opened 2024-01-21
task solution

VeDono:


main.js
```diff
@@ -0,0 +1,9 @@
+'use strict';
+
+const { createServer } = require('./src/createServer');
+
+createServer()
+  .listen(3000, () => {
+    // eslint-disable-next-line no-console
+    console.log('Server is running on localhost:3000');
```
roman-mirzoian, line 8: It is better to keep the port number in a separate variable

src/controllers/expense.controller.js
```diff
@@ -0,0 +1,148 @@
...
+const { getFilteredExpenses } = require('../utils/getFilteredExpenses.js');
+
+const getAll = async(req, res) => {
+  const {
+    userId: recivedUserId,
+    categories,
+    from,
+    to,
+  } = req.query;
+  const userId = Number(recivedUserId);
```
roman-mirzoian, line 14: Why change the name of the user ID, only to later call the variable the same name as it was?

src/controllers/expense.controller.js
```diff
@@ -0,0 +1,148 @@
...
+  }
+
+  await expenseService.remove(expenseToRemove.id);
+
+  res.sendStatus(204);
+};
+
+const patch = async(req, res) => {
+  const { id: recivedId } = req.params;
+  const id = Number(recivedId);
```
roman-mirzoian, line 105:
```suggestion
  const id = Number(req.params.id);
```

src/services/expense.service.js
```diff
@@ -0,0 +1,72 @@
+'use strict';
+
+const { getNewId } = require('../utils/getNewId.js');
+
+const { Expense } = require('../models/Expense.js');
+
+const getAll = () => Expense.findAll();
```
roman-mirzoian, line 7: If we make a request to the server, the function must be asynchronous with the `try...catch` handler

roman-mirzoian (changes requested):
Good job, but there is quite a bit of work to be done.

PR #185, opened 2024-02-01
Solution

Barosz30:


Timi2424 (approved):
Everything works, as I saw yesterday :D But you might want to extract models of users and expenses into seperate files

src/services/expense.service.js
```diff
@@ -0,0 +1,87 @@
...
+const { Expense } = require('../database');
+
+const getAllExpenses = async(req) => {
+  const { userId, categories, from, to } = req.query;
+  const expensesX = await Expense.findAll();
+
+  let filteredExpenses = [...expensesX];
+
+  if (categories && userId) {
+    filteredExpenses = expensesX.filter(item => {
```
Centmsn, line 12: I would strongly recommend to let the database filter the collection (it's gonna be waaay more efficient). At this point, the entire filter logic is on the server.

We can achieve it like this:
```javascript
const whereClause = {};

if(userId) {
  whereClause.userId = userId
}

// and so on...
```

And then, at the very end
```javascript
Expense.findAll({ where: whereClause });
```

src/services/expense.service.js
```diff
@@ -0,0 +1,87 @@
...
+        where: { id }, returning: true,
+      }
+    );
+
+    if (rowsUpdated === 0) {
+      throw new Error('Expenses not found or no changes were made.');
+    }
+
+    return updatedExpense;
+  } catch (error) {
```
Centmsn, line 70: I see no reason to wrap the function code in a `try...catch` block if we're just rethrowing the error without any handling or logging. The outcome would be the same without it.

Centmsn (changes requested):
The code looks pretty good 👍 

I suggested only a few small changes

Centmsn (approved):
Perfect 👌

PR #187, opened 2024-02-02
solution

Timi2424:


src/controllers/expenses.controller.js
```diff
@@ -0,0 +1,146 @@
...
+    console.log(error);
+    res.status(500).send('Internal Server Error');
+  }
+};
+
+const getOne = async(req, res) => {
+  const { id } = req.params;
+
+  if (!id) {
+    res.status(400).send('Bad Request: Missing required parameter');
```
Centmsn, line 29: make sure to use `return` keyword here in order to prevent rest of the function from running.

src/models/expenses.model.js
```diff
@@ -0,0 +1,42 @@
+'use strict';
+
+const { DataTypes } = require('sequelize');
+const { sequelize } = require('../utils/db');
+
+const Expenses = sequelize.define('Expenses', {
+  id: {
+    type: DataTypes.INTEGER,
```
Centmsn, line 8: I would strongly recommend to use `DataType.UUID` for id like values

src/models/user.model.js
```diff
@@ -0,0 +1,21 @@
+'use strict';
+
+const { sequelize } = require('../utils/db');
+const { DataTypes } = require('sequelize');
+
+const User = sequelize.define('User', {
+  id: {
+    type: DataTypes.INTEGER,
```
Centmsn, line 8: Just like above - `UUID` is a better choice here

src/services/expenses.service.js
```diff
@@ -0,0 +1,67 @@
+'use strict';
+
+const Expenses = require('../models/expenses.model');
+
+const getAllExpenses = async(userId, categories, from, to) => {
+  let expenses = await Expenses.findAll();
+
+  if (categories && userId) {
+    expenses = expenses.filter(item => item.category === categories);
```
Centmsn, line 9: It appears that we are implementing filtering logic on the server rather than allowing the database to handle it for us. It is generally more efficient to retrieve from the database only the values that are actually needed.

You could approach it like that:
```javascript
const whereClause = {};

if(userId) {
  whereClause.userId = userId
}

// and so on...
```

and then, at the very end
```javascript
Expense.findAll({ where: whereClause });
```

src/services/users.service.js
```diff
@@ -0,0 +1,40 @@
...
+  return User.findAll();
+};
+
+const getUsersById = async(id) => {
+  return User.findByPk(id);
+};
+
+const createUser = async(name) => {
+  const user = {
+    id: Math.floor(Math.random() * 10000),
```
Centmsn, line 15: Let's add default value to the model itself, instead of generating some random number.

In the user model add:
```javascript
  id: {
      type: DataTypes.UUID,  // already exists
      defaultValue: DataTypes.UUIDV4,
      primaryKey: true // already exists
  }
  ```

Centmsn (changes requested):
I left you a few comments to slightly improve the code 😄

danon321 (approved):

PR #189, opened 2024-02-04
task solution

thesidorenko:


src/controllers/expense.controller.js
```diff
@@ -0,0 +1,78 @@
...
+
+const update = async(request, response) => {
+  const { id } = request.params;
+
+  if (!request.body) {
+    return response.sendStatus(400);
+  }
+
+  // eslint-disable-next-line max-len
+  const expense = await expenseService.updateExpenseById(parseInt(id), request.body);
```
DarkMistyRoom, line 51: better to break the line

src/controllers/expense.controller.js
```diff
@@ -0,0 +1,78 @@
...
+    request.query.userId = parseInt(request.query.userId);
+  }
+
+  const expenses = await expenseService.getExpenses(request.query);
+
+  response.send(expenses);
+};
+
+const getOne = async(request, response) => {
+  const { id } = request.params;
```
DarkMistyRoom, line 16: also would be great to check if `id` was provided

DarkMistyRoom: same for all analogical cases

DarkMistyRoom (approved):
Good job! Left a few minor comments

PR #195, opened 2024-02-05
Solution

mosowiecka:


mbruhler (changes requested):
Your app should be organized in controllers, services and routes. In that shape I can't accept your PR.

PR #198, opened 2024-02-15
add task solution

MishaLehotskyi:


src/controllers/user.controller.js
```diff
@@ -0,0 +1,73 @@
...
+    res.sendStatus(404);
+
+    return;
+  }
+
+  res.status(200).json(user);
+};
+
+const create = async(req, res) => {
+  const body = req.body;
```
l4st1m0za, line 23: Consider using object destructuring to access the name property of req.body. This would make your code more concise and easier to read.

src/utils/createWhereObj.js
```diff
@@ -0,0 +1,39 @@
+'use strict';
+
+const { Op } = require('sequelize');
+const createWhereObj = (params) => {
+  const obj = { where: {} };
+  const { userId, categories, from, to } = params;
+
+  if (userId) {
+    obj.where.userId = +userId;
```
l4st1m0za, line 9: There is no need to convert userId to number, it should be validated in the controller.

l4st1m0za (approved):
Good job!

PR #208, opened 2024-03-16
add task soution

Softjey:


Softjey:
image
My solution complete all test on locale machine, tests failed on github, because i use ecma 2022 syntax in my solution

src/modules/Expense/Expense.service.js
```diff
@@ -0,0 +1,68 @@
...
+      expense, {
+        where: { id }, returning: true,
+      },
+    );
+
+    if (updatedExpense.length === 0) {
+      throw new Error(404);
+    }
+
+    return updatedExpense[0];
```
Andriy-Shymkiv, line 64: I think that updating an entity [this](https://sequelize.org/docs/v7/querying/update/#updating-a-row-using-modelupdate)  way is more safer

src/modules/User/User.service.js
```diff
@@ -0,0 +1,55 @@
...
+      {
+        where: { id }, returning: true,
+      },
+    );
+
+    if (updatedUsers.length === 0) {
+      throw new Error(404);
+    }
+
+    return updatedUsers[0];
```
Andriy-Shymkiv, line 51: same as with expenses

src/utils/Router.js
```diff
@@ -0,0 +1,21 @@
...
+    this.router = express.Router();
+    middlewares.forEach(middleware => this.router.use(middleware));
+
+    routes.forEach(route => {
+      const { method, path, handler } = route;
+
+      this.router[method](path, handler.bind(controller));
+    });
+  }
+}
```
Andriy-Shymkiv, line 19: this looks buggy to be honest
consider to do this:
```suggestion
class Router {
  constructor({ controller, middlewares, routes }) {
    this.router = express.Router();
    middlewares.forEach(middleware => this.router.use(middleware));

    routes.forEach(route => {
      const { method, path, handler } = route;

      this.router[method](path, handler.bind(controller));
    });
  }
}
```

and then usage:
```js
const router = new Router({
  controller: {},
  middlewares: [],
  routes: {}
})
```

Andriy-Shymkiv (changes requested):
looks very good!
left a few comments

src/modules/Expense/Expense.service.js
Softjey: `async update(expenseToUpdate) {
    const { id } = expenseToUpdate;
    const expense = await this.Expense.findByPk(id);
    const updatedExpense = await expense.update(expense);

    return Object.assign(expense, updatedExpense);
  }`
  
  
  Like that?

SanyaBratashchuk (approved):

PR #210, opened 2024-03-17
Add solution

Xilston:


.env
DanilWeda, line 1: add env to git ignore files

DanilWeda (approved):
Good job. Approved. 
Please, do not forget about adding env to gitignore. 
Many thanks)

PR #211, opened 2024-03-18
Solution

pivkopa:


src/db.js
```diff
@@ -15,18 +15,13 @@ const {
...
-*/
-
 const sequelize = new Sequelize({
   database: POSTGRES_DB || 'postgres',
   username: POSTGRES_USER || 'postgres',
   host: POSTGRES_HOST || 'localhost',
   dialect: 'postgres',
   port: POSTGRES_PORT || 5432,
-  password: POSTGRES_PASSWORD || '123',
+  password: POSTGRES_PASSWORD || '123123',
```
Xilston, line 24: great one :)
it's ok for this context, but it would be better to use `.env` file

Xilston (approved):
Good job!

DanilWeda (approved):
Great! 
Approved. 
Thanks)

PR #222, opened 2024-04-20
Solved

DenisGurskiy:


src/controllers/expense.controller.js
```diff
@@ -0,0 +1,115 @@
+const expensesService = require('../services/expense.service.js');
+const userService = require('../services/user.service');
+
+const getExpenses = async (req, res) => {
+  const options = {};
+
+  if (req.query.from) {
```
loralevitska, line 7: Destructure values from `req.query`

loralevitska (changes requested):
Almost done) Let's improve:

PR #224, opened 2024-04-29
add task solution

vk-workshop:


src/controllers/users.controller.js
```diff
@@ -0,0 +1,78 @@
+const usersService = require('../services/users.service');
+
+const getAll = async (req, res) => {
+  res.status(200).send(await usersService.getAll());
+};
+
+const getOne = async (req, res) => {
+  const { id } = req.params;
```
DarkMistyRoom, line 8: also would be great to check if `id` was passed

src/controllers/users.controller.js
```diff
@@ -0,0 +1,78 @@
...
+    return;
+  }
+
+  const user = await usersService.create(name);
+
+  res.status(201).send(user);
+};
+
+const update = async (req, res) => {
+  const { id } = req.params;
```
DarkMistyRoom, line 35: the same for `id`

DarkMistyRoom: validate it everywhere

DarkMistyRoom (approved):
Great job! Check the comments

PR #225, opened 2024-05-03
solution

HolyRaveOS:


HolyRaveOS:
I tested all functionality with postman

src/controllers/expense.controller.js
```diff
@@ -0,0 +1,99 @@
...
+
+const get = async (req, res) => {
+  const { userId, categories, from, to } = req.query;
+
+  const expenses = await expensesService.getExpenses(
+    userId,
+    categories,
+    from,
+    to,
+  );
```
BudnikOleksii, line 12:
```suggestion
const get = async (req, res) => {
  const expenses = await expensesService.getExpenses(req.query);
```
If you do nothing with variables that you get with destructing you don't need to do it

src/controllers/expense.controller.js
```diff
@@ -0,0 +1,99 @@
...
+  }
+
+  const expense = await expensesService.create(
+    userId,
+    spentAt,
+    title,
+    amount,
+    category,
+    note,
+  );
```
BudnikOleksii, line 52: The same, just pass `req.body`

src/helpers/expense.helpers.js
```diff
@@ -0,0 +1,78 @@
+const userService = require('../services/user.service');
+const expenseService = require('../services/expense.service');
+
+const isUserExist = async (userId, res) => {
+  if (!(await userService.getById(userId))) {
+    res.status(400).send('User not found');
+
+    return true;
+  }
```
BudnikOleksii, line 9: Helpers should know anything about the response and what status/message to send. Just check if the user exists

Fix logic in all helpers

src/helpers/expense.helpers.js
```diff
@@ -0,0 +1,78 @@
...
+  return {
+    id,
+    userId,
+    spentAt,
+    title,
+    amount,
+    category,
+    note,
+  };
+};
```
BudnikOleksii, line 71: What is the point of this function?
If you want to remove some values from the object you can use the rest operator.
```javascript
const normalize = (user) => {
  const { password, ...rest } = user;

  return rest;
};
```

src/helpers/user.helpers.js
```diff
@@ -0,0 +1,35 @@
...
+
+  return false;
+};
+
+const normalize = ({ id, name }) => {
+  return {
+    id,
+    name,
+  };
+};
```
BudnikOleksii, line 29: The same about helpers

src/helpers/getRandomNumberFromUUID.js
```diff
@@ -0,0 +1,13 @@
...
+  const uuidBuffer = uuidv4(null, Buffer.allocUnsafe(16));
+
+  const randomNumber = uuidBuffer.readUInt32BE(0);
+
+  return randomNumber;
+}
+
+module.exports = {
+  getRandomNumberFromUUID,
+};
```
BudnikOleksii, line 13: You don't need to generate id manually, database can do it itself if you set default value to column or set to autoincrement

src/models/Expense.model.js
```diff
@@ -1,9 +1,45 @@
...
 const { sequelize } = require('../db.js');
 
 const Expense = sequelize.define(
-  // your code goes here
+  'Expense',
+  {
+    id: {
+      type: DataTypes.INTEGER,
+      primaryKey: true,
+    },
```
BudnikOleksii, line 12: Set `autoIncrement` and you don't have to generate ids manually
https://sequelize.org/docs/v6/core-concepts/model-basics/
image

BudnikOleksii (changes requested):
Good work 👍
Check the suggestions that I've left

src/controllers/expense.controller.js
```diff
@@ -0,0 +1,95 @@
...
+  const expenses = await expensesService.getExpenses(req.query);
+
+  res.send(expenses.map((expense) => expensesHelpers.normalize(expense)));
+};
+
+const getOne = async (req, res) => {
+  const { id } = req.params;
+
+  if (await expensesHelpers.isExpenseExist(id)) {
+    res.status(404).send('Expense with this id not found');
```
Moroz-Dmytro, line 14: Create constants for all status codes

src/controllers/expense.controller.js
```diff
@@ -0,0 +1,95 @@
...
+    return;
+  }
+
+  const expense = await expensesService.getExpenseById(id);
+
+  res.send(expensesHelpers.normalize(expense));
+};
+
+const create = async (req, res) => {
+  const { userId, title, amount, category } = req.body;
```
Moroz-Dmytro, line 25: Destructuring is not really needed

src/helpers/user.helpers.js
```diff
@@ -0,0 +1,25 @@
+const userService = require('../services/user.service');
+const nameCheck = (name) => {
```
Moroz-Dmytro, line 2: Function name should start with a verb

src/helpers/user.helpers.js
```diff
@@ -0,0 +1,25 @@
...
+  if (!name || typeof name !== 'string') {
+    return true;
+  }
```
Moroz-Dmytro, line 5:
```suggestion
  return !name || typeof name !== 'string';
```

Will it work the same?

src/helpers/user.helpers.js
```diff
@@ -0,0 +1,25 @@
...
+};
+
+const isUserExist = async (id) => {
+  if (!(await userService.getById(id))) {
+    return true;
```
Moroz-Dmytro, line 10: Same comments as above

Moroz-Dmytro (changes requested):

BudnikOleksii (approved):
👍

PR #227, opened 2024-05-13
solution

olduma:


src/controllers/expense.controller.js
```diff
@@ -0,0 +1,132 @@
...
+    } = req.body;
+
+    const requestData = {
+      userId,
+      title,
+      amount,
+      category,
+      note,
+      res,
+    };
```
BudnikOleksii, line 26: Do you actually need destructing? You can just pass `req.body`

src/controllers/expense.controller.js
```diff
@@ -0,0 +1,132 @@
...
+const get = async (req, res) => {
+  try {
+    const { userId, categories, from, to } = req.query;
+
+    const expenses = await expensesService.getExpenses({
+      userId,
+      categories,
+      from,
+      to,
+    });
```
BudnikOleksii, line 59:
```suggestion
    const expenses = await expensesService.getExpenses(req.query);
```
The same

src/controllers/expense.controller.js
```diff
@@ -0,0 +1,132 @@
...
+      title,
+      spentAt,
+      category,
+      amount,
+      note,
+    });
+
+    res.status(201).send(expense);
+  } catch (error) {
+    res.status(500).send(INTERNAL_SERVER_ERROR);
```
BudnikOleksii, line 46: - Is it always `500` status and server error? For example, it can be some client error if the data is wrong
- Create a constant with all status codes

src/controllers/expense.controller.js
```diff
@@ -0,0 +1,132 @@
...
+};
+
+const getOne = async (req, res) => {
+  try {
+    const { id } = req.params;
+
+    const expense = await expensesService.getExpenseById(id);
+
+    if (!expense) {
+      return res.status(404).send({ EXPENSE_NOT_FOUND_ERROR });
```
BudnikOleksii, line 74: In this case on the client you see the object where the key and value are the same, probably it should be different. 
```
{
  'Expense not found': 'Expense not found',
}
```
Also in some cases, I see you send just string, standardized your error responses

src/helpers/expense.helpers.js
```diff
@@ -0,0 +1,56 @@
...
+  USER_ID_REQUIRED_ERROR,
+  TITLE_REQUIRED_ERROR,
+  CATEGORY_REQUIRED_ERROR,
+} = require('../utils/config');
+
+const isUserExist = (userId, res) => {
+  const user = userService.getUserById(userId);
+
+  if (!user) {
+    res.status(400).send(USER_NOT_FOUND_ERROR);
```
BudnikOleksii, line 14: - helpers shouldn't know anything about statuses and what response to send, just validate data and throw an error if needed.
- function name should start from verb

src/helpers/expense.helpers.js
```diff
@@ -0,0 +1,56 @@
...
+const validateRequestData = ({
+  userId,
+  title,
+  amount,
+  category,
+  res,
+  isCreatingExpense = true,
+}) => {
+  if (isCreatingExpense && !userId) {
+    res.status(400).send(USER_ID_REQUIRED_ERROR);
```
BudnikOleksii, line 29: The same, fix in all places

src/services/expense.service.js
```diff
@@ -0,0 +1,63 @@
...
+  }
+
+  if (from || to) {
+    conditions.spentAt = {};
+
+    if (from) {
+      conditions.spentAt[Sequelize.Op.gte] = from;
+    }
+
+    if (to) {
```
BudnikOleksii, line 22: Do you need nested `if` statements?

src/services/expense.service.js
```diff
@@ -0,0 +1,63 @@
...
+const create = ({ userId, title, spentAt, category, amount, note }) => {
+  return Expense.create({
+    userId,
+    title,
+    spentAt,
+    category,
+    amount,
+    note,
+  });
+};
```
BudnikOleksii, line 39:
```suggestion
const create = (expense) => {
  return Expense.create(expense);
};
```

src/utils/config.js
```diff
@@ -0,0 +1,11 @@
+module.exports = {
+  INTERNAL_SERVER_ERROR: 'Internal Server Error',
+  USER_NOT_FOUND_ERROR: 'User not found',
+  EXPENSE_NOT_FOUND_ERROR: 'Expense not found',
+  NAME_REQUIRED_ERROR: 'Name is required and must be a string',
+  TITLE_REQUIRED_ERROR: 'Title is required and must be a string',
+  AMOUNT_REQUIRED_ERROR: 'Amount is required and must be a number',
+  CATEGORY_REQUIRED_ERROR: 'Category is required and must be a string',
+  NOTE_REQUIRED_ERROR: 'Note is required and must be a string',
+  USER_ID_REQUIRED_ERROR: 'User ID is required and must be a number',
```
BudnikOleksii, line 10: You can store them in an object:
```javascript
const ERRORS = {
  userNotFound: 'User not found',
  // other errors
};
```

BudnikOleksii (approved):
👍

PR #228, opened 2024-05-15
Node accounting app with db

nk8bnj:


src/controllers/expenses.controller.js
```diff
@@ -0,0 +1,78 @@
...
+  const { id } = req.params;
+  const expense = await expensesService.getExpense(id);
+
+  if (!expense) {
+    res.sendStatus(404);
+
+    return;
+  }
+
+  res.statusCode = 200;
```
BudnikOleksii, line 21: Create an object with status codes to avoid magic numbers

BudnikOleksii (approved):
Good work 👍

PR #229, opened 2024-05-17
Develop

ivanm2706:


src/controllers/controller.user.js
```diff
@@ -0,0 +1,121 @@
+'use strict';
+
+const { User } = require('../models/User.model');
+const { normalizeData } = require('../service/normalize');
+
+async function getAll(req, res) {
+  try {
+    const users = await User.findAll({
+      order: [['name', 'DESC']],
+    });
```
DarkMistyRoom, line 10: it is not a huge mistake but better to move this logic to `services`

src/controllers/controller.expenses.js
```diff
@@ -0,0 +1,204 @@
...
+  }
+
+  if (where[Op.and].length === 0) {
+    delete where[Op.and];
+  }
+
+  req.where = where;
+
+  next();
+}
```
DarkMistyRoom, line 191: let's move the middlewares to a separate file

DarkMistyRoom (approved):
Great job!

PR #231, opened 2024-05-17
Solution

StasivPavlo:


src/controllers/user.controller.js
```diff
@@ -0,0 +1,88 @@
...
+  const userId = req.params.id;
+  const data = matchedData(req, { locations: ['body'] });
+
+  const updatedUser = await userService.updateById(userId, data);
+
+  if (!updatedUser[0]) {
+    return res.status(404).send('Not Found');
+  }
+
+  res.send(await User.findByPk(userId));
```
BudnikOleksii, line 59:
```suggestion
  res.send(await userService.getById(userId));
```
Don't interact with DB directly in the controller. Controllers should interact with services.
In real life, you have even one more layer such as a repository that is responsible for interaction with a DB

BudnikOleksii (approved):
Good work 👍

PR #232, opened 2024-05-17
Develop

RusLan2186:


src/constants/httpStatusCodes.js
```diff
@@ -0,0 +1,10 @@
+const status = {
```
BudnikOleksii, line 1:
```suggestion
const STATUS_CODES = {
```
Constants should be in UPPER_CASE

src/createServer.js
```diff
@@ -1,8 +1,17 @@
...
-};
+const express = require('express');
+const usersRouter = require('./routes/users.route');
+const expensesRouter = require('./routes/expenses.route');
+
+function createServer() {
+  const app = express();
+
+  app.use('/users', express.json(), usersRouter);
+  app.use('/expenses', express.json(), expensesRouter);
```
BudnikOleksii, line 11:
```suggestion
  app.use(express.json());
  
  app.use('/users', usersRouter);
  app.use('/expenses', expensesRouter);
```
You can apply to all routes at once

BudnikOleksii (approved):
Nicely done 🚀

PR #233, opened 2024-05-19
add solution db

tetlisna:


src/services/expenses.service.js
```diff
@@ -0,0 +1,69 @@
...
+  return {
+    id,
+    userId,
+    spentAt,
+    title,
+    amount,
+    category,
+    note,
+  };
+};
```
BudnikOleksii, line 13: What is the point of this function?

src/services/expenses.service.js
```diff
@@ -0,0 +1,69 @@
...
+async function createOne({ userId, spentAt, title, amount, category, note }) {
+  return Expense.create({
+    userId,
+    spentAt,
+    title,
+    amount,
+    category,
+    note,
+  });
+}
```
BudnikOleksii, line 34:
```suggestion
async function createOne(expense) {
  return Expense.create(expense);
}
```
Looks like you don't need destructing

src/services/expenses.service.js
```diff
@@ -0,0 +1,69 @@
...
+) {
+  await Expense.update(
+    {
+      userId,
+      spentAt,
+      title,
+      amount,
+      category,
+      note,
+    },
```
BudnikOleksii, line 48: The same

BudnikOleksii (changes requested):
Good work 👍 
Let's improve your solution slightly

src/constants.js
```diff
@@ -0,0 +1,13 @@
+const OK = 200;
+const CREATED = 201;
+const NO_CONTENT = 204;
+const BAD_REQUEST = 400;
+const NOT_FOUND = 404;
```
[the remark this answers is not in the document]
BudnikOleksii: > You can store them in object
> 
> ```js
> const STATUS_CODES = {
>   ok: 200,
>   // other
> };
> ```

⬆️

src/services/expenses.service.js
BudnikOleksii: > The same

Not fixed

BudnikOleksii (approved):
👍

PR #235, opened 2024-05-20
Add task

ViktoriiaRepo:


src/controllers/expenses.controllers.js
```diff
@@ -0,0 +1,75 @@
...
+
+  res.send(expenses);
+};
+
+const getOne = async (req, res) => {
+  const { id } = req.params;
+  const expense = await expensesService.getById(id);
+
+  if (!expense) {
+    res.sendStatus(404);
```
BudnikOleksii, line 15: Let's create a constant with status codes to avoid magic numbers

src/controllers/expenses.controllers.js
```diff
@@ -0,0 +1,75 @@
...
+const remove = async (req, res) => {
+  const { id } = req.params;
+
+  try {
+    await expensesService.remove(id);
+
+    res.sendStatus(204);
+  } catch (error) {
+    res.sendStatus(404);
+  }
```
BudnikOleksii, line 49: What if other errors occur instead of not found?

src/controllers/user.controllers.js
```diff
@@ -0,0 +1,84 @@
...
+    res.sendStatus(400);
+
+    return;
+  }
+
+  const user = await userService.create(name);
+
+  res.statusCode = 201;
+
+  res.status(201).send(user);
```
BudnikOleksii, line 36: You set status code twice

BudnikOleksii (changes requested):
Looks good 👍 
Just a few fixes are required

BudnikOleksii (approved):
👍

PR #246, opened 2024-06-04
done

2pasha:
test: are tests working?

### DIFF
--- a/src/controllers/user.controller.js
+++ b/src/controllers/user.controller.js
@@ -1,0 +1,75 @@
+const userServices = require('../services/user.service.js');
+
+const get = async (req, res) => {
+  res.send(await userServices.getAll());
+};
+
+const create = async (req, res) => {
+  const { name } = req.body;
+
+  if (!name) {
+    res.sendStatus(400);
+  }
+
+  const user = await userServices.create(name);
+
+  res.statusCode = 201;
+  res.send(user);
+};
+
+const getOne = async (req, res) => {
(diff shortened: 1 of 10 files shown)


src/controllers/user.controller.js
```diff
@@ -0,0 +1,75 @@
...
+  }
+
+  const user = await userServices.create(name);
+
+  res.statusCode = 201;
+  res.send(user);
+};
+
+const getOne = async (req, res) => {
+  const { id } = req.params;
```
DarkMistyRoom, line 21: check if `id` was passed

src/controllers/user.controller.js
```diff
@@ -0,0 +1,75 @@
...
+  const user = await userServices.getById(id);
```
DarkMistyRoom, line 22: also would be great to handle errors on this statge

DarkMistyRoom: applicable to all service requests

src/controllers/user.controller.js
```diff
@@ -0,0 +1,75 @@
...
+    return;
+  }
+
+  await userServices.remove(id);
+  res.sendStatus(204);
+};
+
+const update = async (req, res) => {
+  const { id } = req.params;
+  const { name } = req.body;
```
DarkMistyRoom, line 48: it is better to validate received data immediately after receiving

DarkMistyRoom (approved):
Good job! Just check the comments

PR #252, opened 2024-06-20
done

JustinNowoczek:


choeqq (changes requested):
pleas make sure to divide this files into: 
- controllers
- models
- services
- routes

PR #264, opened 2024-07-11
adding task solution

eduvinagre:


src/controllers/expenseController.js
```diff
@@ -0,0 +1,117 @@
...
+}
+
+async function getAllExpenses(req, res) {
+  try {
+    const userId = req.query.userId;
+    const category = req.query.categories;
+    const filters = {};
+
+    if (userId) {
+      filters.userId = userId;
```
pedro-ruas, line 45: Probably not necessary, as you could directly use:
`filters.userId = req.query.userId`
Or better:
```
const { userId, categories: category } = req.query;
const expeses = await expenseService.getAllExpenses({category, userId});
``` 

Doing so, filters will have undefined values for the missing entries, and so they can later be ignored without the if testing

src/controllers/userController.js
```diff
@@ -0,0 +1,90 @@
+'use strict';
+
+const userService = require('../services/userService');
+
+async function createUser(req, res) {
+  const { name } = req.body;
+
+  if (!name) {
+    return res.status(400).send('Name and  are required');
```
pedro-ruas, line 9: Name and what?

src/controllers/userController.js
```diff
@@ -0,0 +1,90 @@
...
+    res.status(500).send('Error fetching users');
+  }
+}
+
+async function getUser(req, res) {
+  try {
+    const users = await userService.getAllUsers({
+      id: parseInt(req.params.id),
+    });
+    const user = users[0];
```
pedro-ruas, line 36: Could use the same syntax as for expenses:
`const [user] = await userService...`

src/controllers/userController.js
```diff
@@ -0,0 +1,90 @@
...
+  } catch (error) {
+    res.status(500).send('Error fetching user');
+  }
+}
+
+async function updateUser(req, res) {
+  const { name } = req.body;
+
+  if (!name) {
+    return res.status(400).send('Name and  are required');
```
pedro-ruas, line 52: Name and what?

- This test is the same as seen before, maybe you could store it in a validation function?

src/services/expenseService.js
```diff
@@ -0,0 +1,64 @@
...
+        spentAt,
+        title,
+        amount,
+        category,
+        note,
+      });
+
+      return newExpense;
+    } catch (error) {
+      throw error;
```
pedro-ruas, line 19: If what you catch does is only throwing the error, you might get rid of the whole `try/catch`

src/services/expenseService.js
```diff
@@ -0,0 +1,64 @@
...
+
+  async getAllExpenses(filters = {}) {
+    try {
+      const expenses = await models.Expense.findAll({
+        where: filters,
+      });
+
+      return expenses;
+    } catch (error) {
+      throw error;
```
pedro-ruas, line 31: Same here and for all following uses of this

src/services/userService.js
```diff
@@ -0,0 +1,48 @@
...
+
+    if (filters.id) {
+      where.id = filters.id;
+    }
+
+    if (filters.name) {
+      where.name = filters.name;
+    }
+
+    const users = await User.findAll({ where });
```
pedro-ruas, line 23: Seens simply replacing `filters = {}` with `where = {}` and removing the _ifs_ would get the same result

pedro-ruas (changes requested):

src/controllers/expenseController.js
eduvinagre: You're absolutely right. Just did the pointed changes and updated de expenseServices to ensure consistency on the getAllExpenses  method so that the class is explicitly handling userId and category. Let me know if you don't think this wasn't necessary also.

eduvinagre:
@pedro-ruas Please let me know if I could get the ideas in your correction right. Cheers, Mate!

src/controllers/userController.js
```diff
@@ -0,0 +1,99 @@
...
+    return false;
+  }
+
+  return true;
+}
+
+async function createUser(req, res) {
+  const { name } = req.body;
+
+  if (!validateName(name, res)) {
```
pedro-ruas, line 18: I like the idea, but this isn't the best approach.
When working with this type of validation, you can add validators to "intercept" the request before it reaches the controller, so you don't need to return a boolean for the controller to check and prevent the continuation, the validator would pause the operation before the controller. How to do it:
To declare a validator, your function `validateName` is almost  as it should be, but it needs in fact 3 arguments: req, res, next, like this:
```
function validateName(req, res, next) {
    const { name } = req.body; // same as in your controller, because it is the SAME req the controller receives

    if (!name) {
        return res.status(400)... // same code you did in your function. 
    }

    return next(); // this tells Express to continue to the next element in the flow (your controller)
}
```
Now that your validator is declared, add it to your route:
```
 router.post('/',
    validateName,
    userController.createUser
);
```
Now, if the validation fails, the controller won't even know the request has arrived, and you don't have to "double-check" the fails (first check: validation, second check: the boolean return of the validation)

src/services/expenseService.js
```diff
@@ -0,0 +1,56 @@
...
+      title,
+      amount,
+      category,
+      note,
+    });
+
+    return newExpense;
+  }
+
+  async getAllExpenses({ userId, category }) {
```
pedro-ruas, line 19: Intead of `({ userId, category })`, you can use `(where = {})`
Ignore the ifs and them: `const expenses = await models.Expense.findAll({ where })`

src/services/userService.js
```diff
@@ -0,0 +1,39 @@
...
+const { User } = require('../models/User.model');
+
+const userService = {
+  async createUser({ name }) {
+    const newUser = await User.create({ name });
+
+    return newUser;
+  },
+
+  async getAllUsers(filters = {}) {
```
pedro-ruas, line 12: Same here: `getAllUsers(where = {})`, and remove next `where` declaration, as it's not necessary

pedro-ruas (changes requested):
Very well done! Minor improvements still

eduvinagre:
Ok @pedro-ruas , did everything you asked plus had to refactor some other parts of the code as services and models had to be fixed also for the code to work. I added the validators folder with the nameValidation inside it. Hope you like it :)

src/controllers/expenseController.js
```diff
@@ -0,0 +1,123 @@
...
+
+async function createExpense(req, res) {
+  try {
+    const { userId, spentAt, title, amount, category, note } = req.body;
+
+    if (!spentAt || !title || !amount || !userId) {
+      return res.status(400).send('All fields are required');
+    }
+
+    const user = await userService.getAllUsers({ id: userId });
```
pedro-ruas, line 17: Not super needed, but `userService` could have a function called `validadeUser` or something on this matter, that is called to check if that user exists and returns a boolean.
Because you use this test here in multiple places, would be a good idea.

src/controllers/expenseController.js
```diff
@@ -0,0 +1,123 @@
...
+const { ExpenseService } = require('../services/expenseService');
+const expenseService = new ExpenseService();
+const { Op } = require('sequelize');
+const userService = require('../services/userService');
+
+async function createExpense(req, res) {
+  try {
+    const { userId, spentAt, title, amount, category, note } = req.body;
+
+    if (!spentAt || !title || !amount || !userId) {
```
pedro-ruas, line 13: I see you created a validator for another request. So, this could also use validators

pedro-ruas (approved):
Great improvement!! 🥳

PR #269, opened 2024-07-11
Develop

ekateriina7:


src/controllers/expense.controller.js
```diff
@@ -0,0 +1,157 @@
...
+    filteredExpenses = filteredExpenses.reduce((acc, expense) => {
+      if (categories.includes(expense.category)) {
+        acc.push(expense);
+      }
+
+      return acc;
+    }, []);
+  }
+
+  return filteredExpenses;
```
SanyaBratashchuk, line 55: Controllers are just the entry point for requests and are responsible for sending back the response to the client. Services, on the other hand, contain the business logic. So move business logic to services

src/services/user.service.js
```diff
@@ -0,0 +1,57 @@
+'use strict';
+
+const { User } = require('../models/User.model');
+
+const normalize = ({ id, name }) => {
+  return { id, name };
+};
```
SanyaBratashchuk, line 7: why do you need this function

SanyaBratashchuk (changes requested):

SanyaBratashchuk (approved):

PR #270, opened 2024-07-11
solution

MatheusStefan:


src/controllers/expensesController.js
```diff
@@ -0,0 +1,94 @@
...
+  } catch {
+    res.status(500).send('Internal Server Error');
+  }
+};
+
+const updateExpenseById = async (req, res) => {
+  const { id } = req.params;
+
+  if (!id) {
+    return res.sendStatus(404);
```
pedro-ruas, line 73: I believe this status should be 400 (bad request) and not 404 (not found)

src/controllers/usersController.js
```diff
@@ -0,0 +1,95 @@
...
+    const newUser = await userServices.createUser(name);
+
+    res.status(201).json(newUser);
+  } catch (error) {
+    res.status(500).send('Internal Server Error');
+  }
+};
+
+const deleteUser = async (req, res) => {
+  const { id } = req.params;
```
pedro-ruas, line 45: Probably it is a good idea to check if the id was sent.
By blocking the request if the user did not send all that was needed, you can prevent the db interaction before the fail and save some processing power and money

src/controllers/usersController.js
```diff
@@ -0,0 +1,95 @@
...
+const updateUser = async (req, res) => {
+  const { id } = req.params;
+  const { name, spentAt, title, amount, category, note } = req.body;
+
+  if (!id || !name) {
+    return res.sendStatus(400);
+  }
+
+  try {
+    const userToUpdate = await userServices.updateUserById({
```
pedro-ruas, line 68: That is OK, but if only what you need to test is `id` and `name` you don't need to restructure everything else:
```
const { id, name, ...userProps } = req.body;

if (...)

const userToUpdate = await userServices.updateUserById({ id: Number(id), name, ...userProps });
```

pedro-ruas (changes requested):
Minimal improvements, great job!

src/controllers/expensesController.js
MatheusStefan: The test says it is expect to be 404

jv-aquino (approved):
awesome job Matheus!

PR #272, opened 2024-07-12
add task solution

PrytulaAlexandr:
add task solution

### DIFF
--- a/src/routes/expense.router.js
+++ b/src/routes/expense.router.js
@@ -1,0 +1,15 @@
+const express = require('express');
+const expensesController = require('../constrollers/expenses.controller.js');
+const expensesRouter = express.Router();
+
+expensesRouter.get('/', expensesController.getAllExpenses);
+
+expensesRouter.get('/:id', expensesController.getOneExpense);
+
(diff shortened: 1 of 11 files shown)


src/routes/expense.router.js
```diff
@@ -0,0 +1,15 @@
+const express = require('express');
+const expensesController = require('../constrollers/expenses.controller.js');
+const expensesRouter = express.Router();
+
+expensesRouter.get('/', expensesController.getAllExpenses);
```
DanilWeda, line 5: You can move all routes to constants

src/routes/users.router.js
```diff
@@ -0,0 +1,15 @@
+const express = require('express');
+const userController = require('../constrollers/users.controller.js');
+const usersRouter = express.Router();
+
+usersRouter.get('/', userController.getAllUsers);
```
DanilWeda, line 5: The same, you can add routes to constants

DanilWeda (changes requested):
Great job. 
Check some small minor comments about using constants.
Many thanks)

PR #274, opened 2024-07-16
Create CRUD functionality with DB

olafchuszno:


src/index.js
```diff
@@ -3,6 +3,13 @@
 'use strict';
 
 const { createServer } = require('./createServer');
+const { Expense } = require('./models/Expense.model');
+const { User } = require('./models/User.model');
+
+(async function syncDb() {
+  await User.sync({ force: true });
+  await Expense.sync({ force: true });
+})();
```
Zibi95, line 12: Add error handling to asynchronous operations to catch any potential errors. Also it would be good to just console.log if db has succesfully synced. You could also sync whole `db` by calling `sync` method on `sequlize` instance.

src/index.js
```diff
@@ -3,6 +3,13 @@
...
 
 createServer().listen(5700, () => {
```
Zibi95, line 14: Keep such values that configure server in env file.

src/createServer.js
```diff
@@ -1,7 +1,47 @@
...
+  // Expenses
+  app.get('/expenses', express.json(), getExpenses);
+
+  app.post('/expenses', express.json(), postExpense);
+
+  app.get('/expenses/:expenseId', express.json(), getExpense);
+
+  app.delete('/expenses/:expenseId', express.json(), deleteExpense);
+
+  app.patch('/expenses/:expenseId', express.json(), patchExpense);
```
Zibi95, line 42: Move express.json() middleware outside individual route handlers and apply it globally to the app.  `app.use(express.json());`

Zibi95, line 42: Use Express Router for modular route handling to keep the main server file clean.
Example
```
const expensesRoute = express.Router();

expensesRoute.use('/:id', validateId);

expensesRoute
  .get('/', getAllExpenses)
  .get('/:id', getExpense)
  .post('/', validateBody, createNewExpense)
  .delete('/:id', deleteExpense)
  .patch('/:id', validateBody, updateExpense);
```

and then in this file with this one line `app.use('/expenses', expensesRoute);` you can connect server with the route

src/controllers/expenses-controller.js
```diff
@@ -0,0 +1,136 @@
...
+};
+
+const deleteExpense = async (req, res) => {
+  const { expenseId } = req.params;
+
+  if (expenseId === undefined) {
+    res.sendStatus(400);
+
+    return;
+  }
```
Zibi95, line 99: Those lines are reapeated like 3 times in this file. Its greate case to create middleware to check if `expenseId` is valid before hitting the controller

src/controllers/expenses-controller.js
```diff
@@ -0,0 +1,136 @@
...
+  } catch {
+    res.sendStatus(404);
+  }
+};
+
+const getExpense = async (req, res) => {
+  const { expenseId } = req.params;
+
+  if (expenseId === undefined) {
+    res.sendStatus(400);
```
Zibi95, line 58: Add detailed error messages.  
`return res.status(400).send({ error: 'Bad Request: Missing expenseId' });`

src/controllers/expenses-controller.js
```diff
@@ -0,0 +1,136 @@
...
+
+    return;
+  }
+
+  try {
+    const newExpense = await createExpense(req.body);
+
+    res.status(201).send(newExpense);
+  } catch {
+    res.sendStatus(400);
```
Zibi95, line 88: This one should be 500 error bcs BE server has some internal error in this case

src/controllers/users-controller.js
```diff
@@ -0,0 +1,84 @@
...
+  const newUser = await createUser(name);
+
+  res.statusCode = 201;
+  res.send(newUser);
+};
+
+const getUser = async (req, res) => {
+  const userId = +req.params.userId;
+
+  const searchedUser = await getUserById(userId);
```
Zibi95, line 32: What about 500 internal error. You should use try catch blocks on async actions.

src/models/Expense.model.js
Zibi95, line 1: If there are relationships between User and Expense (e.g., one-to-many), consider defining associations using Sequelize's association methods.

for example
```
User.hasMany(Expense, { foreignKey: 'userId' });
Expense.belongsTo(User, { foreignKey: 'userId' });
```

This way you can reduce the amount of manual SQL you need to write, making your code cleaner and more maintainable.

here is example of fetching a user along with their expenses.
 
```
const usersWithExpenses = await User.findAll({
  include: Expense,
});
```

src/services/ExpenseService.js
```diff
@@ -0,0 +1,87 @@
...
+  const newExpense = await Expense.create({
+    userId: +userId,
+    spentAt,
+    title,
+    amount: +amount,
+    category,
+    note,
+  });
+
+  return newExpense;
```
Zibi95, line 41: remember to use try and catch blocks for async operations

Zibi95 (changes requested):
Well done. I left some comments below. There is some work to do to improve this code further.  What I also suggest is to create FE for this and make you own app. This way you will have fullstack project under your belt that you can show in CV.

Zibi95 (approved):

Zibi95:
I missclicked the approve option 👍

PR #298, opened 2024-09-25
solution

kapesha:


src/controllers/expenceController.js
```diff
@@ -0,0 +1,88 @@
...
+const userService = require('../services/userService.js');
+
+const getExpenses = async (req, res) => {
+  const { userId, categories, from, to } = req.query;
+
+  res.send(await expenseService.getAllExpenses(userId, categories, from, to));
+};
+
+const getOneExpense = async (req, res) => {
+  const { id } = req.params;
```
anastasiiavorobiova, line 11: Consider wrapping the code into a try-catch block and sending 500 errors in the case of db issues

src/controllers/expenceController.js
```diff
@@ -0,0 +1,88 @@
...
+    return;
+  }
+  res.status(200);
+  res.json(expense);
+};
+
+const createExpance = async (req, res) => {
+  const { userId, spentAt, title, amount, category, note } = req.body;
+
+  if (!userId || !spentAt || !title || !amount) {
```
anastasiiavorobiova, line 26: Consider moving such checks to the middleware

anastasiiavorobiova (changes requested):
Great work! Consider fixing the code in the controllers.

SanyaBratashchuk (approved):

PR #309, opened 2024-10-29
Develop

arthurfujii:


src/controllers/expense.controller.js
```diff
@@ -0,0 +1,69 @@
+const expenseService = require('../services/expense.service');
+const { getById: getUserById } = require('../services/user.service');
+
+const get = async (req, res) => {
+  const expenses = await expenseService.getAll(req.query);
+
+  res.send(expenses.map((expense) => expenseService.normalize(expense)));
```
pedro-ruas, line 7: This line can be merged with line 5 and condensed:
```
const expenses = await expenseService.getAll(req.query)
    .then((exs) => exs.map(expenseService.normalize);

res.send(expenses);
```

src/controllers/expense.controller.js
```diff
@@ -0,0 +1,69 @@
...
+};
+
+const getOne = async (req, res) => {
+  const expense = await req.entry;
+
+  res.send(expenseService.normalize(expense));
```
pedro-ruas, line 13: Same advice as before, probably valid for all queries

src/middlewares/validadeExpense.js
```diff
@@ -0,0 +1,17 @@
+const expenseService = require('./../services/expense.service');
+
+const validateExpense = () => {
```
pedro-ruas, line 3: A better naming would be: `validateExpenseExists`

src/middlewares/validateEntry.js
```diff
@@ -0,0 +1,23 @@
+const expenseService = require('./../services/expense.service');
+const userService = require('./../services/user.service');
+
+const validateEntry = (entryType) => {
```
pedro-ruas, line 4: Same, `validateEntryExists`
Just so it is perfectly clear what this function should do and when to be used.
`validateEntry`, can sound like some function that would be used also for creating an entry. However, this one validates only if an entry already exists, thus `validateEntryExists`

src/middlewares/validateFields.js
```diff
@@ -0,0 +1,19 @@
+function validateFields(fields) {
```
pedro-ruas, line 1: `validateRequiredFields` 😉

src/middlewares/validateFields.js
```diff
@@ -0,0 +1,19 @@
...
+  return function (req, res, next) {
+    const extractedValues = {};
```
pedro-ruas, line 3: No need to change, this is good. But here is another way to do it:
```
const extractedValues = fields.reduce((result, field) => ({
    ...result,
    [field]: req.body[field]
}), {});
```

src/middlewares/validateUser.js
```diff
@@ -0,0 +1,17 @@
+const userService = require('./../services/user.service');
+
+const validateUser = () => {
```
pedro-ruas, line 3: `validateUserExists`

src/services/expense.service.js
```diff
@@ -0,0 +1,63 @@
...
+    id,
+    note,
+    spentAt,
+    title,
+    userId,
+  };
+};
+
+const getAll = async (filterParams = {}) => {
+  const { categories, from, to, ...query } = filterParams;
```
pedro-ruas, line 16: This destructuring can be done in the function declaration:
```
const getAll = async ({ categories, from, to, ...query } = {}) => { ...
```
This will have the same effect

src/utils/generateId.js
```diff
@@ -0,0 +1,9 @@
+const generateId = function (array) {
+  if (array.length > 0) {
```
pedro-ruas, line 2: Can also be done like this:
```
const generateId = function (array) {
    return Number((array.at(-1) ?? { id: -1 }).id) + 1;
}
```
This tries to get the last element from the array `array.charAt(-1)`. If the array is empty, this method returns `undefined`.
When `undefined`, the fallback returns an object with the id "-1": `?? { id: -1 }`.
Then, it takes the id from that result (either the id from the last element, either "-1".
After that, it converts this id to a number `Number(...)`.
And finally, adds 1 to it `+ 1`.

If the array is empty, an object with id = -1 will be returned, -1 with + 1 returns 0 for the initial id of the database, and adds as should for all others

pedro-ruas (changes requested):
Minimal improvements, mostly in naming

pedro-ruas (approved):
Niiica!

PR #312, opened 2024-11-08
Develop

v-zagorovskii:


src/controllers/expense.controller.js
```diff
@@ -0,0 +1,112 @@
...
+
+const createExpense = async (req, res) => {
+  const data = req.body;
+
+  if (
+    !data.userId ||
+    !data.title ||
+    !typeof data.categories ||
+    !typeof data.spentAt ||
+    !typeof data.amount
```
SanyaBratashchuk, line 44: typeof always returns string so it will be truthy anyway

src/services/expense.service.js
```diff
@@ -0,0 +1,74 @@
...
+    if (from) {
+      params.spentAt[Op.gte] = from;
+    }
+
+    if (to) {
+      params.spentAt[Op.lte] = to;
+    }
+  }
+
+  const expList = Expense.findAll({ where: params });
```
SanyaBratashchuk, line 31:
```suggestion
  const expList = await Expense.findAll({ where: params });
```

SanyaBratashchuk (changes requested):

SanyaBratashchuk (approved):

PR #315, opened 2024-11-28
Develop

italomagno:


src/controllers/expenses.controller.js
```diff
@@ -0,0 +1,154 @@
...
+    }
+
+    res.status(500).json({
+      message: error.message,
+    });
+  }
+};
+
+const createNewExpense = async (req, res) => {
+  console.log('body: r,', req.body);
```
joaorpereira, line 55: remove console.log()

src/models/Expense.model.js
```diff
@@ -1,9 +1,46 @@
 'use strict';
 
 const { sequelize } = require('../db.js');
+// const { User } = require('./User.model.js');
```
joaorpereira, line 4: if it's not been used, let's remove it

joaorpereira (approved):

PR #339, opened 2025-02-02
node_accounting-app-with-db

Abdrakhmanov-Renat:


anastasiiavorobiova (changes requested):
Great work! 
- Consider making validations and sending the 400 error before accessing the controllers
- Consider wrapping the whole code inside the controller into the try-catch block and sending the 500 error if needed (sometimes database issues happen and your code should be stable in this case)
- If there are no filtered expenses you can send an empty array

PR #385, opened 2025-05-09
Develop

kovaltar:
Unfortunately, in this task, 10 tests were failing because the tests, for some reason, do not account for cascade deletion when a foreign key is defined in the models.
I manually checked each of the failing tests — all cases worked correctly.
Removing the foreign key from the model just to pass the tests is incorrect and does not comply with the standards of relational (linked) tables.

Therefore, I made the decision to replace one line in the user.tests.js file, specifically:
the line:
await User.destroy({ truncate: true });
was replaced with:
await User.destroy({ where: {}, truncate: true, cascade: true });

This way, the data from the previous test is correctly deleted before the next test runs, and each test uses its own data — as originally intended.

### DIFF
--- a/config/config.json
+++ b/config/config.json
@@ -1,0 +1,23 @@
+{
+  "development": {
+    "username": "postgres",
+    "password": "P0$tgre$123",
+    "database": "postgres",
+    "host": "127.0.0.1",
+    "dialect": "postgres"
+  },
+  "test": {
+    "username": "postgres",
+    "password": "P0$tgre$123",
+    "database": "postgres",
+    "host": "127.0.0.1",
+    "dialect": "postgres"
+  },
+  "production": {
+    "username": "postgres",
+    "password": "P0$tgre$123",
+    "database": "postgres",
+    "host": "127.0.0.1",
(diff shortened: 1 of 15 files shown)


kovaltar:
@mateacademy-ai-mentor regarding these cases:

    Issue: If Expense.findByPk(id) returns null (expense not found), calling expense.destroy() will throw an error. Consider checking if the expense exists before calling .destroy(), and handle the case where it does not exist.

    Issue: If Expense.findByPk(id) returns null, calling expense.update(data) will throw an error. Add a check to ensure the expense is not null before calling .update().

    Issue: If User.findByPk(id) returns null (user not found), calling user.destroy() will throw an error. Consider checking if the user exists before calling .destroy(), and handle the case where it does not exist.

    Issue: If User.findByPk(id) returns null, calling user.update(data) will throw an error. Add a check to ensure the user is not null before calling .update().

I already have a check for the existence of the user before calling the deleteExpense, updateExpense, deleteUser, and updateUser functions in the corresponding controllers.

anastasiiavorobiova (approved):
Great work!

PR #393, opened 2025-05-28
Solution

msoares96:
I had to fix the tests because they weren't handling the "before each" cycle correctly. I did manual test using postman and everything worked perfectly. The final version drops the foreign key settings and required "category" field so the tests pass.


pedro-ruas (approved):
Great job! And I'm aware of the problems in these tests, you did great!

PR #457, opened 2025-12-01
done task

nbryk:


nbryk:
@mateacademy-ai-mentor thank you for your review, but all test passed. The code works

PR #469, opened 2026-01-13
solution

nadiaaamf:
Accounting app (with Node.js and PostgreSQL)


pedro-ruas (changes requested):
This work looks great! There is just one thing I want you to change.
The file `createServer.js` has too much code on it.
In a real world project, this file would be divided in modules, in a way that it makes sense.
For example, you should split it in at least 2 new files: `user.controller.js` and `expense.controller.js`. Each file should then handle their respective routes.

pedro-ruas (approved):
Perfect! Now it looks professional

PR #471, opened 2026-01-15
Solution

hma-3:


hma-3:
@mateacademy-ai-mentor I did the task according to tests and [documentation](https://app.swaggerhub.com/apis/YURIIHOLIUK_1/Accounting-App/1.0.0). Both of them describe collections `Users` and `Expenses`. There is no `Category` model. But I updated the relationships between `User` and `Expense` according to CRUD

PR #493, opened 2026-03-11
Add Expense model with user association and CRUD routes

marcellealvb:


raulriato (changes requested):
Don't throw the code entirelu in just one file. Separate it in folders to make it readable and easier to maintain. Make at least a Controller and a Service.

IanLuan (changes requested):
Great job separating the files, but the tests are failing. Please run `npm run test` and fix the issues so all tests pass.

PR #510, opened 2026-04-02
v1

wzfxc:


vadiimvooo (changes requested):
Good job, but you have to pass all tests. Feel free to ask for help in fe_chat.
Consider why the test "should create a new expense if optional fields is not provided" fails with a 400 error, even though the test name suggests category is optional—could the model or controller be requiring it as mandatory?
Look at the test "should return all expenses by category": what query parameter does it use for filtering, and does it match what the service expects?
In the update method of the controller: what is returned in the response—the old object before the update or the new one after? Why might this affect the test?

brespect (approved):
Well done! 👍

PR #522, opened 2026-04-22
app-with-db

Vlshedevr:


src/createServer.js
```diff
@@ -1,7 +1,23 @@
...
+const expensesRouter = require('./routes/expenses.routes');
+const categoriesRouter = require('./routes/categories.routes');
+
 const createServer = () => {
-  // your code goes here
+  const app = express();
+
+  app.use(cors());
+
+  app.use('/users', express.json(), userRouter);
```
Denys-Kravchuk9988: I would recommend to use `app.use(express.json());` instead of duplicating in every endpoint. Also there is a sense to add additional improvement (not required but it's used in real projects)
`app.use(express.json({ limit: '1mb' }));`

Denys-Kravchuk9988 (changes requested):
Good job! 

Added 1 comment and please try to resolve comments from AI

brespect (changes requested):
Good progress, but you need to pass all tests before requesting the review

brespect (approved):
Well done!

PR #554, opened 2026-07-08
node_accounting-app-with-db

Elber-Kevenny:


src/userController.js
```diff
@@ -0,0 +1,245 @@
...
+  const postExpense = async (request, response) => {
+    const { spentAt, title, category, note } = request.body;
+    const userId = parseInt(request.body.userId);
+    const amount = parseInt(request.body.amount);
+
+    if (!spentAt || !title || !amount || !userId) {
+      return response.sendStatus(400);
+    }
+
+    const isUser = User.findOne({ where: { id: userId } });
```
DanielCamposDevX: Missing await for database promise

src/userController.js
```diff
@@ -0,0 +1,245 @@
...
+    if (from && to) {
+      filtros.spentAt = { [Op.between]: [from, to] };
+    }
+
+    try {
+      const allExpense = await Expense.findAll({
+        where: filtros,
+      });
+
+      if (allExpense.length === 0) {
```
DanielCamposDevX: This condition runs thee same code if true or false

src/userController.js
```diff
@@ -0,0 +1,245 @@
...
+      }
+      await User.destroy({ where: { id: id } });
+      response.sendStatus(204);
+    } catch (error) {
+      // eslint-disable-next-line no-console
+      console.error(error, 'erro');
+
+      response.sendStatus(404);
+
+      return response.status(500).json({ error: error });
```
DanielCamposDevX: response will be always 404, as its sending, response.sendStatus(404) beforee the 500

src/userController.js
```diff
@@ -0,0 +1,245 @@
...
+      // eslint-disable-next-line no-console
+      console.error('Erro');
+
+      return response.status(500).json({ error: erro });
+    }
+  };
+
+  const postAddUser = async (request, response) => {
+    try {
+      const { name } = await request.body;
```
DanielCamposDevX: request.body is not a promise, it does not need to be awaited.

src/userController.js
```diff
@@ -0,0 +1,245 @@
...
+    } catch (error) {
+      // eslint-disable-next-line no-console
+      console.error('Erro');
+
+      return response.status(500).json({ error: error });
+    }
+  };
+
+  const getOne = async (request, response) => {
+    const id = await parseInt(request.params.id);
```
DanielCamposDevX: parseInt is a syncronous function, it does not need to be awaited.

DanielCamposDevX (changes requested):
Add awai to user validation.
Remove 404 status error from catch block on deleteUser.
*comments on the code *

src/controllers/expensesController.js
raulriato, line 1: The name of the file is `userController`, so it should deal with `user` requests only. Put the `expenses` requests in `expensesController`

src/userController.js
```diff
@@ -0,0 +1,236 @@
...
+const createUserController = () => {
+  const { User, Expense } = models;
+
+  const getUser = async (request, response) => {
+    try {
+      const result = await User.findAll();
+
+      response.status(200).json(result);
+    } catch (erro) {
+      // eslint-disable-next-line no-console
```
raulriato: You should not disable `eslint`. Instead, your code should be written to work with it

src/userController.js
```diff
@@ -0,0 +1,236 @@
...
+    }
+  };
+
+  const postAddUser = async (request, response) => {
+    try {
+      const { name } = request.body;
+
+      if (!name) {
+        return response.sendStatus(400);
+      }
```
raulriato: Only if you decide to practice implementing a services layer, it would be nice to add some validation to ensure that the name is valid—for example, checking that it doesn't contain numbers or invalid special characters.

If you decide not to implement this layer, the current validation is enough for this task.

raulriato (changes requested):
You are mixing different organizational patterns in your project.

You have separated your models into their own folder and created different files for users and expenses. This is correct, and I recommend following the same approach for your controllers and routers.

Currently, you have a single router file handling both users and expenses. Additionally, your userController is also handling expense-related logic.

I recommend following the same pattern you are already using for your models:

src/
├── controllers/
│   ├── expensesController.js
│   └── usersController.js
├── models/
│   ├── expensesModel.js
│   └── usersModel.js
└── routers/
    ├── expensesRouter.js
    └── usersRouter.js

Also, note that the file names in the example follow a consistent naming convention. Currently, you use names such as User.model.js and userController.js, which follow different conventions.

It doesn't matter which naming convention you choose, as long as you choose one and use it consistently throughout the project.

If you want an extra challenge, you could add another layer to the project: services.

A services layer is not necessary for this project because it is still relatively small. However, it becomes useful in larger projects, as it helps separate business logic from the controllers. Implementing it here could be a good opportunity to practice this type of architecture.

With this layer, the controllers would be responsible for handling the requests and responses, while the services would contain the business logic, such as validations and data processing.

raulriato (approved):
Good job!

The `models` files are still with the names different from the others, they use `singular` instead of `plural` (`user` instead of `users`). But since it's no major problem, I'll approve it.

